PR #17, opened 2026-06-06
Prefer Claude Keychain credentials for quota lookup

### LINKED ISSUE #16: Claude quota lookup can prefer stale file credentials over Keychain credentials
## Context

Tokcat can show a Claude OAuth error in the Agent limits card even when Claude Code itself is authenticated and has a valid credential in macOS Keychain.

> The failure is limited to the quota lookup path. Local token history still reads from local usage logs and is not part of this issue.

## Observed Behavior

| Check | Result |
|---|---|
| `~/.claude/.credentials.json` | Exists and contains `claudeAiOauth`, but may be an older credential shape without fields such as `expiresAt` and `scopes` |
| macOS Keychain item `Claude Code-credentials` | Contains a newer Claude Code OAuth credential with expiry and scopes |
| Tokcat credential selection | Reads the file credential first and returns before checking Keychain |
| User-visible result | Claude Agent limits can show an OAuth token error even though Claude Code is logged in |

## Suspected Cause

The Claude credential loader currently checks sources in this order:

```mermaid
flowchart TD
    ENV[Environment override] --> FILE[~/.claude/.credentials.json]
    FILE --> KC[macOS Keychain]
    FILE --> STOP[Returns stale file credential when file exists]
```

On current Claude Code installs, Keychain appears to be the fresher credential source. If the legacy file exists, Tokcat never reaches Keychain, so the usage request can be made with a stale or less complete credential.

## Expected Behavior

Tokcat should prefer the active Claude Code credential source on macOS while keeping the existing environment override and file fallback behavior.

```mermaid
flowchart TD
    ENV[Environment override] --> KC[macOS Keychain]
    KC --> FILE[~/.claude/.credentials.json fallback]
```

## Notes

| Constraint | Detail |
|---|---|
| Scope | Claude OAuth quota lookup only |
| Privacy | No token values need to be logged or exposed |
| Compatibility | The file credential can remain as fallback for older setups or non-Keychain environments |

Nanako0129:
## Summary

This changes the Claude OAuth credential lookup order used by the Agent limits quota request.

| Before | After |
|---|---|
| Environment override → `~/.claude/.credentials.json` → macOS Keychain | Environment override → macOS Keychain → `~/.claude/.credentials.json` |

Fixes #16.

## Why

On macOS, Claude Code can have a fresh OAuth credential in Keychain while a legacy `~/.claude/.credentials.json` file still exists. Tokcat previously returned the file credential first, so the quota lookup could use a stale credential and show a Claude OAuth error even though Claude Code itself was authenticated.

> This only changes the Agent limits OAuth quota lookup. Local usage history still comes from local session logs.

## Verification

| Check | Result |
|---|---|
| `cargo check` | Passed |
| Local app install smoke test | Built and launched `/Applications/Tokcat.app` successfully during manual validation |

### DIFF
--- a/src-tauri/src/agent_usage.rs
+++ b/src-tauri/src/agent_usage.rs
@@ -468,6 +468,10 @@ fn load_claude_credentials() -> Result<ClaudeCredentials, String> {
         return Ok(credentials);
     }
 
+    if let Some(raw) = load_claude_credentials_from_keychain()? {
+        return parse_claude_credentials_data(&raw);
+    }
+
     let credentials_path = claude_credentials_path();
     match fs::read_to_string(&credentials_path) {
         Ok(raw) => return parse_claude_credentials_data(&raw),
@@ -479,10 +483,6 @@ fn load_claude_credentials() -> Result<ClaudeCredentials, String> {
                 error
             ));
         }
-    }
-
-    if let Some(raw) = load_claude_credentials_from_keychain()? {
-        return parse_claude_credentials_data(&raw);
     }
 
     Err("Claude OAuth credentials not found. Run `claude` to authenticate.".to_string())


handlecusion:
I've reviewed it. Thank you for the contribution!

PR #27, opened 2026-07-10
Add Grok Build usage history, live tail, and OAuth quota

oesni:
## Summary

Add **Grok Build** as a first-class client so Tokcat can show local token history, live throughput, and OAuth quota limits alongside Claude Code, Codex, and the other supported tools.

Inspired by the log and quota shapes documented in [tokscale](https://github.com/junhoyeo/tokscale) (`sessions/grok`, usage/grok), implemented natively in Tokcat’s Rust reader (no runtime dependency on the tokscale CLI).

### Local usage history
- Scan `$GROK_HOME/sessions/*/*/updates.jsonl` (default `~/.grok/sessions`)
- Turn deltas from cumulative `totalTokens` (input bucket; no stable I/O split in Grok logs)
- Reconcile compaction remainders via sibling `signals.json`
- Model metadata from summary / update meta / events fallback
- Dedup keys include workspace path so the same session id under different cwds does not collide
- Bundled xAI/Grok pricing (including `grok-composer-*` aliases)

### Live session
- Tail `updates.jsonl` for tokens/min (same tailer used for Claude/Codex/Hermes)
- Monotonic high-water handling for transient counter dips

### Agent limits
- Gate on `auth.json` presence under `$GROK_HOME` / `~/.grok`
- Fetch billing (gRPC-web), task usage (Frequent/Occasional), and subscription plan
- Prefer install paths for the `grok` binary when using the optional agent billing fallback
- Plan-only responses still show identity without forcing an error status

### UI / docs
- Client style + icon, Live trace label, Agent limits rows
- README (EN/KO), landing/llms copy: list Grok Build and note Grok OAuth quota network calls

## Out of scope
- Side-by-side local “Tokcat Dev” install helpers
- Mapping API tier enums to marketing plan names (e.g. SuperGrok Heavy)
- Full OAuth token refresh (same as tokscale’s Grok usage client: uses current access token)

## Verification

| Check | Result |
|---|---|
| `cargo test --lib` (Grok + agent_usage related) | Passed |
| Diff scan for secrets / absolute home paths / real emails | Only test fixtures (`primary@example.com`, dummy tokens) |
| Manual smoke | Local Grok sessions surface in history; Agent limits shows plan/task windows when credentialed |

## Notes

- Usage history stays local. Grok quota lookups call `grok.com` only when local OAuth credentials already exist (same product pattern as Codex/Claude limits).
- SuperGrok subscription spend is still estimated with API list prices for multi-client cost charts.

Screenshot 2026-07-11 at 3 44 19 AM


handlecusion (approved):
Reviewed the full diff (+2146/-63, 14 files) with a focus on the OAuth/quota network surface. LGTM — approving.

## Security / privacy — clean
- All network targets are official xAI hosts only (`auth.x.ai`, `grok.com/rest/*`, `grok.com/grok_api_v2.*`). No third-party exfiltration path.
- Credentials are read from local `~/.grok/auth.json` and sent as `Bearer` only to those hosts. Tokens are not logged (error strings carry status/index only, never the token).
- Everything is gated on `auth.json` presence (`grok_is_configured`), so uninstalled users make zero network calls — consistent with the existing Codex/Claude behavior and the disclosure added to the README/landing copy.
- The optional `grok agent --no-leader stdio` fallback spawns the user's own resolved `grok` binary (GROK_HOME/bin → ~/.grok/bin → PATH), 4s timeout then killed — same shape as the existing `claude --version` spawn. Acceptable.

## Correctness — solid, well tested
- Monotonic high-water handling ignores counter dips without lowering `last_total`, so recovering totals aren't double-counted (both live tail and graph).
- Per-turn deltas + compaction reconciliation via `signals.json` + workspace-scoped dedup keys (avoids same-session-id collisions across cwds and re-scan duplicates). Traced logic matches the test expectations.
- Pricing branch ordering (`grok-4-5` → `grok-4-1` → `grok-4` catch-all) and `grok-composer-*` alias handling are correct.

## Verification
- `cargo test --lib grok` → **17/17 passed**, compiles cleanly (only pre-existing `native_tray.rs` unsafe warnings, unrelated to this PR).

## Non-blocking notes (optional follow-up)
1. Pre-turn baseline tokens (before the first `user_message_chunk`) are excluded from graph history. Looks intentional (tokscale parity; `signals.json` reconciles any compaction remainder) — just confirming the intent.
2. `usage_graph`/`usage_tail` parse `totalTokens` with strict `as_i64()`, so a float-encoded total would be skipped (whereas `agent_usage`'s `grok_numeric_value` also accepts strings/objects). Token counts are integers in practice, so no real impact — minor parser-consistency nit.

Nice work — additive, no regressions, good coverage.

PR #34, opened 2026-07-16
Fix Codex Business usage and Claude OAuth refresh

raphaelmiran:
## Summary

- parse Codex Business `spend_control.individual_limit` usage when `rate_limit` is null
- show the account spend limit as a Monthly spend window, including percentage fallbacks and reset time
- send Claude OAuth refresh requests as JSON with the credential scopes required by the current Claude Code flow
- add regression coverage for both response formats

## Verification

- `cargo test` — 43 passed
- `npx tsc --noEmit`
- `npm run build`
- installed the local macOS app and verified both Codex Business and Claude limits display correctly

### DIFF
--- a/src-tauri/src/agent_usage.rs
+++ b/src-tauri/src/agent_usage.rs
@@ -15,8 +15,6 @@ const CODEX_CLIENT_ID: &str = "app_EMoamEEZ73f0CkXaXp7hrann";
 const CLAUDE_USAGE_URL: &str = "https://api.anthropic.com/api/oauth/usage";
 const CLAUDE_REFRESH_URL: &str = "https://platform.claude.com/v1/oauth/token";
 const CLAUDE_CLIENT_ID: &str = "9d1c250a-e61b-44d9-88ed-5944d1962f5e";
-const CLAUDE_DEFAULT_SCOPES: &str =
-    "user:profile user:inference user:sessions:claude_code user:mcp_servers";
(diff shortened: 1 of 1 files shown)


raphaelmiran:
@handlecusion Codex was not working with Chatgpt business plan because the api replies a diffrerent json format.

handlecusion (commented):
Thanks for this, @raphaelmiran! 🙏

The Codex Business `spend_control.individual_limit` parsing and the form-urlencoded → JSON switch for the Claude refresh are both spot on. The token endpoint rejects the old urlencoded body, so the JSON change is the real refresh fix.

I pushed one follow-up commit (beda2df) that drops the `scope` parameter from the refresh request:

- The canonical Claude Code refresh flow (and other working clients, e.g. opencode) send only `grant_type` / `refresh_token` / `client_id` — no `scope`. The JSON body switch alone restores refresh; the `scope` field was incidental.
- Per RFC 6749 §6, omitting `scope` on a `refresh_token` grant makes the server reissue the **originally granted** scopes, which is exactly what we want.
- The `CLAUDE_DEFAULT_SCOPES` fallback (`user:sessions:claude_code`, `user:mcp_servers`) is unverified and would be sent whenever the stored credentials carry no scopes — risking an `invalid_scope` error that breaks refresh entirely.

Everything else LGTM — `cargo test` passes (16 tests), build is clean. Merging after this. Thanks again for the fix! 🎉

Issue #22, opened 2026-06-22
No Cursor usage detected


kyle-erikson:
Just installed a fresh tokcat via Homebrew. Upon launch of the application, it picked up Claude Code and Codex. However, I do not have Codex installed, nor a subscription to it. I do have Cursor installed, both IDE and CLI, but neither are detected. Any thoughts?

handlecusion:
Thanks for the report, @kyle-erikson — you've actually surfaced two separate things, and both are on us.

**1. Codex showing up even though it's not installed**
The live usage panel currently hardcodes two agents (Claude + Codex) and renders a tile for each regardless of whether you've authenticated. With no Codex credentials present, that tile falls back to an error / "sign in" state instead of being hidden — so it looks "detected" when it really isn't. Fix: hide (or clearly mark as not-configured) agents you haven't set up.

**2. Cursor not detected**
This is a real gap. Right now Cursor usage is only read from a legacy cache file (`~/.config/tokscale/cursor-cache/usage*.csv`) that older Tokcat/tokscale setups generated. A fresh Homebrew install never has that file, and we don't yet read Cursor's native local usage data — so Cursor shows nothing even though it's installed (both IDE and CLI). Fix: add a native Cursor parser so a clean install picks it up automatically.

Both fixes are planned and I'm picking this up now — I'll update here when they ship in a release.

z0ph:
Hello,

Does this fix apply to Cursor IDE version 3.10? I've installed tokcat, but Cursor usage isn't displayed; only token or cost is shown. I'm more interested in the percentage of usage against my monthly plan to control my spend.

z0ph:
First of all, thanks a lot for this lightning-fast update. Appreciated!

I've tried the new version, and it now shows the usage, but I have some concerns with the displayed number; see the screenshot below.

<img width="2738" height="1652" alt="Image" src="https://github.com/user-attachments/assets/6598d713-9a97-4eb9-8c36-a9371bed849a" />

My goal is to have visibility across the current usage of the plan or token for my ChatGPT, Claude (Desktop), and Cursor subscriptions in a single tool.

PS: Thanks a lot for making this piece of software and maintaining it as OSS.

handlecusion:
@z0ph — fixed in **[0.1.41](https://github.com/handlecusion/tokcat/releases/tag/v0.1.41)** (`brew upgrade --cask tokcat`). Your screenshot pinned it down exactly, thank you.

[...]

PR #43, opened 2026-08-04
Refresh agent quota cards every 30 minutes

z0ph:
Follow-up to #39 / #41 — from my side of #22. The Cursor pool numbers those PRs added are correct; they just stop updating after launch.

## The problem

`agent_usage::run()` — the call that hits `GetCurrentPeriodUsage` and builds the Cursor pool windows — only runs when the frontend invokes `get_agent_usage`, and the frontend only invokes it when the key passed to `useAgentUsage` changes:

- `useAgentUsage` re-fetches on its `refreshKey` dep alone (`src/hooks/useAgentUsage.ts`).
- `refreshTick` is bumped by the header refresh button, ⌘R, the tray *Refresh Now* item, and the Cursor-usage toggle — nothing else (`src/App.tsx`).
- The main webview is created hidden at launch and lives for the whole session, so the mount fetch happens exactly once per launch.

The 30-minute loop in `spawn_refresh_loop` (`src-tauri/src/lib.rs`, `REFRESH_SECS = 1800`) rebuilds `usage_graph` and, when the toggle is on, `cursor_usage::refresh_cache`. It never touches `agent_usage`. The one 30-minute `setInterval` already in `App.tsx` belongs to `checkForUpdatesSilent`.

Net effect with **Settings → Menubar → Plan usage → Cursor**: the percentage is pinned to launch time. It doesn't look broken either — the tray-title effect keeps re-running because `overviewStats` and `tokensPerMin` change (rate updates land every 3 minutes), so it repaints a stale number indefinitely. Opening the popover doesn't help; the `popover-shown` listener only resizes the window. On a menubar app that stays running for days, the displayed quota can be days old.

This isn't Cursor-specific. Codex, Claude and Grok Build tiles come from the same `agent_usage::run()` and are equally frozen — Cursor is just where it became visible, because it's the pool that moves fastest during a work session.

## Changes

A 30-minute tick for agent usage, matching `REFRESH_SECS` so the quota tiles and the graph payload age at the same rate.

The tick bumps `agentUsageTick`, kept separate from `refreshTick` and summed into the hook's key. Folding it into `refreshTick` would have been a shorter diff, but it would also spin the header cat every half hour and re-invoke `refresh_graph`, duplicating the rebuild the backend loop already runs.

Cost is at most four upstream calls per 30 minutes, each already gated on that client being configured.

## Alternative considered

Driving this from Rust inside `spawn_refresh_loop` and emitting an `agent-usage-update` event for `useAgentUsage` to listen on, the way `useGraphStream` consumes `graph-update`. That's the better architectural fit and keeps the cadence defined in one place, but it's a much larger diff and needs a guard so the mount-time invoke and the loop don't double-fetch. Happy to redo it that way if you'd rather the cadence live in the backend.

## Not in this PR

Re-fetching on `popover-shown` behind a short TTL, so opening the dashboard shows current numbers rather than up-to-30-minute-old ones. That's a UX call rather than a bug fix, so I left it out — say the word if you want it folded in.

## Tests

Verified against the dev server (`npm run dev`), which drives the same hook through `/api/agent-usage`, with the interval temporarily set to 3s and network traffic captured in DevTools:

- Before (`main`): 1 × `/api/agent-usage` for the entire session, no matter how long the page stayed open.
- After: repeated `/api/agent-usage` on the interval, while `/api/rate` and `/api/stream` were still fetched exactly once — so nothing else re-fetches as collateral.

Interval restored to 30 minutes before committing. `tsc --noEmit` and `npm run build` clean. No Rust changes, so no `cargo` runs.

### DIFF
--- a/README.ko-KR.md
+++ b/README.ko-KR.md
@@ -113,7 +113,7 @@ Settings에서 두 가지 스타일 선택 가능: 회전하는 고양이 또는
 |------|------|
 | **2D / 3D 사용량 보기** | 최근 30일 누적 토큰 차트 또는 full-year 3D 타일 그래프. orbit 컨트롤, 카메라 상태 저장, active 타일 자동 fit. |
 | **Overview + 클라이언트 탭** | 데이터가 있는 Claude Code, Codex CLI, Cursor IDE, OpenCode, Gemini CLI, Copilot CLI, Amp, Droid, Hermes, Grok Build 탭으로 전환. |
-| **Agent limits** | 로컬 credential이 있으면 Codex/Claude/Grok/Cursor의 session, weekly, monthly, credits, model, reset, remaining limit window 표시. Cursor는 청구 주기의 모델 풀별(Cursor Models, Other Models)로 한 줄씩 보여주며, Cursor 자체 Plan & Usage 화면과 동일한 수치입니다. |
+| **Agent limits** | 로컬 credential이 있으면 Codex/Claude/Grok/Cursor의 session, weekly, monthly, credits, model, reset, remaining limit window 표시. Cursor는 청구 주기의 모델 풀별(Cursor Models, Other Models)로 한 줄씩 보여주며, Cursor 자체 Plan & Usage 화면과 동일한 수치입니다. 한도 값은 실행 시, 이후 30분마다, 그리고 Refresh Now를 누를 때 다시 조회합니다. |
 | **라이브 메뉴바 타이틀** | 오늘의 토큰 / 오늘의 비용 / 전체 토큰 / 전체 비용 / 아이콘만. 토큰 처리량 업데이트는 3분마다 전달됩니다. |
 | **트레이 아이콘 애니메이션** | 회전하는 고양이 또는 party parrot 애니메이션 속도가 실시간 토큰 처리량에 따라 변동. |
 | **네이티브 비브런시 + 글래스모피즘** | 투명 윈도우 + macOS `sidebar` 머티리얼; 라이트/다크 자동 전환. |
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ Pick between two styles in Settings: the spinning cat or a party parrot. During 
 |---------|---------|
 | **2D / 3D usage views** | Recent 30-day stacked token bars or interactive full-year 3D tile graph with orbit controls, persistent camera, and auto-fit-to-active-tiles framing. |
 | **Overview + client tabs** | Switch between all-client totals and dedicated tabs for Claude Code, Codex CLI, Cursor IDE, OpenCode, Gemini CLI, Copilot CLI, Amp, Droid, Hermes, and Grok Build when data is present. |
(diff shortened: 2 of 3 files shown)


handlecusion:
Diagnosis checks out. Confirmed independently that `useAgentUsage` depends on `refreshKey` alone (`src/hooks/useAgentUsage.ts:46`) and that `spawn_refresh_loop` (`src-tauri/src/lib.rs:243`) only touches `cursor_usage::refresh_cache` and `usage_graph::run` — `agent_usage` is never on that path. Keeping this off `refreshTick` was the right call: the header cat stays still and `refresh_graph` is not re-invoked.

`tsc --noEmit` and `npm run build` are clean here too. On the dev server I reproduced your result — `/api/agent-usage` repeats on the interval while `/api/rate` and `/api/stream` are each fetched exactly once, so nothing re-fetches as collateral.

## Verified in the real hidden webview

The dev-server test leaves open the one question that decides whether this works in production: the window is created with `"visible": false` and stays hidden, and macOS suspends JS timers in a hidden `WKWebView`. So I ran a dev build with an `eprintln!` probe in `get_agent_usage`, window hidden, popover never opened:

| `AGENT_USAGE_REFRESH_MS` | observed gaps |
|---|---|
| 15s | 181s, 182s, 182s, 182s |
| 300s | 364s, 364s, 364s |

**It works, but not for the reason the code suggests.** Timers in the hidden webview only fire when the `rate-update` event on `RATE_EMIT_SECS` (`src-tauri/src/lib.rs:21`, 180s) wakes it — hence the 182s quantum. The 300s row is the useful one: it fires at 364s (2 × 182), not 728s, so the wall clock does keep running and delivery just rounds up to the next wake.

For 30 minutes that means `ceil(1800 / 182) × 182 = 1820s` — fires at ~30m20s. Well within tolerance, so **the fix does what it says on a real menubar session.**

The catch is that it holds by coincidence of two constants that never reference each other: if the rate loop's cadence ever changes or it stops emitting, this tick degrades silently back to the exact bug you are fixing. I filed #44 with the measurements rather than expanding scope here — and it is a concrete argument for the backend-driven approach you sketched under "Alternative considered", since a `tokio` interval is immune to webview suspension.

## Minor, non-blocking

- `refreshTick + agentUsageTick`: correct today since both counters are monotonic, but summing two keys is fragile — if either is ever reset, the sum can collide and a re-fetch is silently skipped. `useAgentUsage(refreshTick, agentUsageTick)` or a single `agentUsageKey` that both paths bump would be sturdier.
- The tick runs even when the tray mode is not `plan_percent` and the popover has never been opened. Worth noting the cost is a little above "four upstream calls": `claude_is_configured()` shells out to the Keychain, and a token refresh rewrites credentials on disk. `get_agent_usage` also does blocking fs/process work directly in an `async` command without `spawn_blocking` — pre-existing, but this PR makes it recur.
- Nice catch updating both READMEs.
- The `popover-shown` re-fetch you left out is the bigger UX win, since the popover is exactly when anyone reads these numbers. Happy to see it as a follow-up.

Unrelated: `package-lock.json` still says `0.1.38` while `package.json` is `0.1.41`, so `npm install` dirties the tree.

LGTM.

PR #45, opened 2026-08-05
Drive agent quota refresh from the backend loop

### LINKED ISSUE #44: Frontend intervals in a hidden webview only fire when `rate-update` wakes it — undocumented coupling to RATE_EMIT_SECS
The main webview is created with `"visible": false` (`src-tauri/tauri.conf.json`) and stays hidden for most of a menubar session. macOS suspends JS timer *delivery* in a hidden `WKWebView`, so any `setInterval` in `src/App.tsx` only fires when something else wakes the webview.

Today the only thing that reliably wakes it is the `rate-update` event emitted on `RATE_EMIT_SECS` (`src-tauri/src/lib.rs:21`, 180s). That makes every frontend interval implicitly coupled to the rate loop — a dependency that appears nowhere in the code.

## Measured

Instrumented `get_agent_usage` with an `eprintln!` probe, ran a dev build with the window hidden and the popover never opened:

| `setInterval` period | observed firing gaps |
|---|---|
| 15s | 181s, 182s, 182s, 182s |
| 300s | 364s, 364s, 364s |

The wake period is ~182s, matching `RATE_EMIT_SECS = 180` plus overhead.

Two things follow from the 300s row:

- 364s = 2 × 182s, not 728s — the **wall clock keeps running** while suspended, only delivery is deferred. Timers round up to the next wake boundary rather than accumulating drift.
- A sub-wake-period interval collapses to one firing per wake (standard `setInterval` coalescing, no queued catch-up).

## Impact

Nothing is broken right now. The two intervals in `App.tsx` are both 30 minutes, and `ceil(1800 / 182) × 182 = 1820s`, so they fire ~30m20s apart — close enough that #43's quota refresh and `checkForUpdatesSilent` both work as intended.

The problem is that this holds only by coincidence of the constants:

- If the rate loop ever stops, is gated behind a setting, or its cadence grows past 30 minutes, **the agent-usage refresh and the silent update check degrade with it, silently**. There is no failure signal — the tiles just go back to showing launch-time values, which is exactly the #43 symptom.
- Any future frontend interval shorter than ~3 minutes will not do what its constant says it does.
- The comment on `AGENT_USAGE_REFRESH_MS` points at `REFRESH_SECS`, which is not actually what governs when it fires.

## Suggested direction

Drive the cadence from Rust rather than from a webview timer, as @z0ph already sketched in the "Alternative considered" section of #43: refresh agent usage inside `spawn_refresh_loop` and emit an `agent-usage-update` event for `useAgentUsage` to listen on, the way `useGraphStream` consumes `graph-update`. A `tokio` interval is not subject to webview suspension, and it puts the cadence in one place instead of splitting it between a Rust constant and a JS constant that only agree by convention.

Needs a guard so the mount-time `invoke` and the loop's first emit do not double-fetch.

If that is more churn than wanted, the cheaper stopgap is a comment on both `App.tsx` intervals recording that sub-3-minute periods are not achievable in the hidden webview and why.

handlecusion:
Closes #44. Follow-up to #43, which fixed the stale quota tiles with a `setInterval` in `App.tsx`.

## The problem with the timer

That interval works, but not for the reason the code says. The main window is created with `"visible": false` and stays hidden for most of a menubar session, and macOS suspends JS timer delivery in a hidden `WKWebView`. The timer only fires when something else wakes the webview — in practice the `rate-update` event on `RATE_EMIT_SECS` (`src-tauri/src/lib.rs:21`, 180s).

Measured on a hidden window with an `eprintln!` probe in `get_agent_usage`, popover never opened:

| `AGENT_USAGE_REFRESH_MS` | observed gaps |
|---|---|
| 15s | 181s, 182s, 182s, 182s |
| 300s | 364s, 364s, 364s |

The 300s row is the informative one: it fires at 364s (2 × 182), not 728s, so the wall clock keeps running and delivery just rounds up to the next wake. At 30 minutes that lands on `ceil(1800 / 182) × 182 = 1820s`, ~30m20s — fine.

So nothing is broken today. What's broken is the coupling: `AGENT_USAGE_REFRESH_MS` and `checkForUpdatesSilent` both silently depend on the rate loop's cadence, nothing in the code says so, and if that ever changes the tiles go back to showing launch-time values with no error anywhere — the exact #43 symptom.

## Changes

Move the cadence into `spawn_refresh_loop` and emit `agent-usage-update`, consumed by `useAgentUsage` the same way `useGraphStream` consumes `graph-update`.

A `tokio` interval isn't subject to webview suspension, and the emit wakes the webview itself instead of waiting for another subsystem to.

Some things fall out for free:

- the frontend interval and `agentUsageTick` are gone, and with them the `refreshTick + agentUsageTick` summed key that #43 review flagged as fragile
- cadence lives in `REFRESH_SECS` alone, rather than `1800` in Rust and `30 * 60 * 1000` in JS agreeing only by comment
- quota refresh now shares a cycle with the graph rebuild rather than running on a parallel clock

Mount fetch and manual refresh (header button, ⌘R, tray *Refresh Now*) are untouched — `get_agent_usage` stays, and the key is back to `refreshTick` alone.

On the double-fetch guard raised in #44: `spawn_refresh_loop` sleeps before its first iteration, so the first emit is one `REFRESH_SECS` after startup and never races the mount `invoke`. The listener is also registered before the mount fetch, mirroring `useGraphStream`, so a push landing mid-invoke isn't dropped.

## Tests

- `tsc --noEmit`, `npm run build`, `cargo check` all clean.
- Behaviour verified on a real hidden window (dev build, popover never opened) with `REFRESH_SECS = 60` and probes on both sides — a `[EMIT]` in the loop and an `[INVOKE]` fired from the frontend listener:

```
[INVOKE] get_agent_usage from webview   <- mount
[EMIT] agent-usage-update
[INVOKE] get_agent_usage from webview
[EMIT] agent-usage-update
[INVOKE] get_agent_usage from webview
[EMIT] agent-usage-update
[INVOKE] get_agent_usage from webview
```

Every emit reached the frontend immediately. Note the cadence itself is the proof: 60s is well under the 182s wake period, so this is a rate a webview timer cannot achieve at all.

Constants restored before committing.

## Not in this PR

The `popover-shown` re-fetch behind a short TTL, still worth doing — the popover is when anyone actually reads these numbers, and they can be up to 30 minutes old either way.


handlecusion (commented):
Independent review pass on this branch. The direction is right and the frontend diff is fine, but the Rust side has a problem I missed when writing it: awaiting `agent_usage::run()` inline turns a per-tile failure into a whole-loop failure. Two findings worth fixing before merge.

## 1. A hung `agent_usage::run()` now stops the entire refresh loop — HIGH

`src-tauri/src/lib.rs:263`

```rust
let _ = app.emit("agent-usage-update", &agent_usage::run().await);
let years = state.known_years();   // everything below waits on the line above
```

The graph rebuild, the `graph-update` emits and `tray::refresh_tray_title` all sit behind that await, with no `timeout` and no separate task. `run()` (`agent_usage.rs:288`) is sequential across codex → claude → grok → cursor, and while every HTTP call is capped at 30s (`:502`, `:580`, `:668`, `:886`, `:945`) two subprocess calls are **not bounded at all**:

- `agent_usage.rs:857` — `/usr/bin/security find-generic-password`, plain `.output()`. Reached every cycle through `claude_is_configured()`. If the binary's keychain ACL stops matching (a code-signature change after an app update is the usual trigger) `security` blocks on a modal prompt the user never sees, because the window is hidden.
- `agent_usage.rs:2116` — `Command::new("claude").arg("--version").output()`, plain `.output()`, called while building the Claude request header. A wedged npm shim or a stalled PATH entry blocks forever.

Concrete failure: keychain ACL mismatch after an update → from that moment the contribution graph never rebuilds, `graph-update` never fires, the tray title freezes and `cursor_usage::refresh_cache` never runs, until restart. Under #43 the same hang cost only the quota tiles, because `get_agent_usage` ran as its own command task.

Fix is small: `async_runtime::spawn` it as an independent task, and/or wrap in `tokio::time::timeout`. The blocking subprocess work also belongs behind `spawn_blocking` — the loop already does exactly that for `usage_graph::run` two lines down.

In fairness `cursor_usage::refresh_cache().await` (`lib.rs:255`) is already on this path, so there's precedent — but it is opt-in, has a 30s timeout, and spawns no subprocesses. This change deepens an existing wart into a real hazard.

## 2. The cycle is now `1800s + run() duration` — MEDIUM

Same line. Because the emit is awaited inline the period drifts by however long `run()` took. Worst realistic case isn't small: on a captive portal or with dead DNS, four agents × 30s HTTP timeouts plus grok's per-credential billing calls can add minutes to every cycle, and the graph and tray title are frozen for all of it. The PR body claims quota refresh "shares a cycle with the graph rebuild" — true, but the sharing is serial, not concurrent, which was not the intent. Spawning it separately fixes this and finding 1 together.

## 3. Behaviour after system sleep regresses — MEDIUM, needs a decision

`tokio::time::sleep` is backed by `Instant`, which on Apple targets does not advance while the system is asleep. #44's measurement established the opposite for the webview timer: the wall clock kept running during suspension and only delivery was deferred (the 300s → 364s row).

So: lid closes 5 minutes into a cycle, machine sleeps 8 hours, user opens the lid and clicks the menubar icon. Under #43 the JS deadline had long passed and fired at the first wake — fresh within ~3 minutes. Under this PR the tokio sleep still has 25 minutes left on its uptime clock, so the popover shows 8-hour-old quota for another 25 minutes.

That's the exact staleness this PR series exists to remove, arriving at the most likely moment of use for a laptop menubar app. To be clear about evidence: the tokio side follows from the clock source, but the JS-side comparison across a real system sleep is inferred from #44's suspension data rather than measured. Either way it argues for pulling the `popover-shown` re-fetch out of "Not in this PR" — it stops being a nicety and becomes the thing that covers wake-from-sleep.

## Smaller

- **Dev-server path lost its periodic refresh** (`src/hooks/useAgentUsage.ts:15-31`). The non-Tauri branch fetches `/api/agent-usage` once per `refreshKey` and never again; `agentUsageTick` used to re-run both branches. Dev-only, but the PR body should say so.
- **`unlisten` can be dropped** (`useAgentUsage.ts:38-60`). If cleanup runs during the `await import`s or `await listen(...)`, cleanup sees `null` and the handler registered a tick later never gets removed. Bounded — the orphan closure captures its own `disposed = true`, so it can't corrupt state, and there's no `StrictMode` in `src/main.tsx` to amplify it — so it's a leaked IPC handler per unluckily-timed manual refresh. Worth a `cancelled` flag, not worth blocking. Note `useGraphStream` has the identical hole, so "mirrors useGraphStream" is accurate but inherits it.
- **A slow `invoke` can overwrite a newer push** (`useAgentUsage.ts:45-50`). Listening before fetching stops a push being *dropped*, not *lost*: if ⌘R's `get_agent_usage` takes 60s (plausible per finding 1) and a periodic emit lands at second 30, the resolving invoke overwrites the fresher payload with its older snapshot. Cosmetic, self-corrects next cycle.

## Checked and not real

- **Double-fetch at first launch** — `spawn_refresh_loop` sleeps before its first iteration (`lib.rs:250`), so the first emit is at t=1800s and cannot race the mount invoke. The claim in the PR body holds.
- **Error-state regression** — `run()` never returns `Err`; per-agent failures live in `snapshot.error` inside the payload, so the emit path loses nothing. The hook is marginally better than before: a later push clears a stale `error` left by a failed mount fetch.
- **Payload shape mismatch between emit and invoke** — both are `AgentUsagePayload` with the same `serde(rename_all = "camelCase")`, and `get_agent_usage` always returns `Ok`. Identical JSON.
- **`plan_percent` tray title not updating on a hidden window** — the effect needs a React render, and it gets one: the existing `rate-update` listener already drives renders on a hidden window as a shipped feature. The new path is structurally identical.
- **Listener destroyed when the popover closes** — `tray::hide_popover` calls `w.hide()`, not close. The webview and its listeners survive.
- **Dropped event during manual-refresh re-registration** — the effect does re-register per `refreshTick`, leaving a sub-millisecond gap. Against a 30-minute emit cadence that's negligible, and the manual path does its own full fetch anyway.
- **Concurrent OAuth token refresh** between a manual invoke and the loop rewriting `auth.json` — real, but pre-existing: #43's periodic tick was itself an invoke of the same command and raced identically. Not introduced here.

## Bottom line

Findings 1 and 2 are one change — spawn the refresh as its own task with a timeout — and should land before merge. Finding 3 is a real regression this PR's description doesn't acknowledge, and I'd rather fold the `popover-shown` re-fetch in than ship the wake-from-sleep gap.

handlecusion:
Fixed findings 1 and 2 in 7a16b4f.

**The loop no longer waits.** The refresh is spawned as its own task, so the graph rebuild, the `graph-update` emits and the tray title refresh don't sit behind four upstream calls, and the period stays `REFRESH_SECS` instead of `REFRESH_SECS + run()`. A cycle that overruns is superseded by the next one.

That alone would only move the hang off the critical path, not bound it — a wedged cycle would still park a task every 30 minutes for the rest of the session. So the two unbounded subprocess calls are now bounded too: `/usr/bin/security` (`agent_usage.rs:857`) and `claude --version` (`:2116`) went through `Command::output()` with no deadline; they now go through a helper that kills the child at 5s. Every HTTP call in the module was already capped at 30s, so those two were the only holes.

Semantics are preserved: `Err` still means a spawn failure and `Ok(None)` a non-zero exit, with a timeout folded into the latter — which is what both callers already act on.

Four tests cover the helper, including that an overrunning child is killed rather than waited on. 56 Rust tests pass, `tsc --noEmit` and `npm run build` clean.

Not re-run on device after this commit: the keychain read only exercises meaningfully on a machine with Claude credentials, and `.output()` internally does the same pipe-and-read-to-end this helper does, so the path is equivalent. Worth one manual launch before merge to confirm the Claude tile still populates.

Finding 3 (wake-from-sleep) is untouched and still open — it needs a scope decision rather than a fix, so I'd rather settle that separately than bundle it here.

handlecusion:
Finding 3 fixed in 14d8931, by pulling the `popover-shown` re-fetch in rather than deferring it.

`useAgentUsage` now listens for `popover-shown` and re-fetches when the snapshot is older than 60s. Age is measured against the payload's own `generatedAt`, not a local timer, so time spent asleep counts — which is the whole point, since the tokio sleep's clock does not count it.

Not routed through `refreshTick`: no header spin, no graph rebuild, just a top-up. Inside the TTL the listener does nothing, so opening and closing the popover repeatedly costs no upstream calls.

Verified on a hidden window with probes on both the emit site and the command:

```
[INVOKE] get_agent_usage      <- mount
[SHOWN] popover-shown emitted <- opened at ~30s, inside the TTL: no fetch
[SHOWN] popover-shown emitted <- opened past 60s
[INVOKE] get_agent_usage      <- re-fetch
```

Both READMEs updated. 56 Rust tests pass, `tsc --noEmit` and `npm run build` clean.

This closes out all three findings. Remaining known gaps, none blocking:

- the two LOW frontend nits from the review (the `unlisten` race, and a slow `invoke` overwriting a newer push) are untouched — both are bounded and self-correcting
- the non-Tauri dev-server path still has no periodic refresh
- CI runs `cargo check` but not `cargo test`, so the four new helper tests only ran locally. Worth a separate one-line workflow change.

handlecusion:
Second review round. Both HIGH findings are real and both were mine; fixed in a804dd5.

## 1 — `child.wait()` had no deadline, so the guard bounded nothing

Confirmed. `read_to_end` returns the moment the pipe closes, and a child can close stdout and keep running, so the success path parked the caller for the child's full lifetime *and* reported success. The commit that introduced this also spawned the refresh, which made it quieter and therefore worse: the loop stops noticing and starts another next cycle, so tokio workers go one at a time until every `#[tauri::command]` stops responding.

The reap now shares the read's deadline (`wait_until`, polling `try_wait`) and kills on both paths. A test covers it — `/bin/sh -c 'echo hi; exec 1>&-; sleep 30'` with a 300ms deadline, which took 30s before the fix.

## 2 — folding a timeout into `Ok(None)` deletes the Claude card

Confirmed, and the chain is exactly as described: `claude_is_configured` fell through to `claude_credentials_path().exists()`, which is false on a Keychain-backed install, so `run()` never pushed a Claude snapshot and the tile vanished with no error text. That the scenario is precisely the one the timeout was added for makes it worse, not more acceptable.

Outcomes are three-way now (`CommandOutput::Ok` / `Failed` / `TimedOut`). A non-zero exit still means "not there" and falls through; a stall returns an error naming the keychain prompt, and `claude_is_configured` treats an error as configured so the card stays and shows what happened.

## 3, 4, 6 — taken

- In-flight guard on the popover top-up, so overlapping opens during a slow refresh can't stack fetches.
- TTL raised to 5 minutes. You're right that 60s made the 30-minute cadence meaningless for an active session; the top-up only needs to cover what the backend clock can't.
- `apply` now drops payloads older than the newest seen, which fixes both the stale render and `generatedAtRef` walking backwards into a redundant top-up.
- The swallowed rejection now sets `error` while keeping the current payload.

## 5, 7 — acknowledged, not in this PR

`run()` still blocks a worker for up to 3 × 5s across two keychain reads and `claude --version`, and `claude_is_configured` still throws away credentials it just fetched. Both want `run()` behind `spawn_blocking` plus a single credential read, which is a wider change than this PR should carry — filing separately. Same for the Codex `fs::write` token race, which is narrow and pre-existing but genuinely more reachable now that two schedulers can call `run()`.

## 3 (leaked reader thread) — accepted, now documented

You're right that `kill()` signals the child alone, so a grandchild holding the pipe leaves the reader thread blocked. Killing the process group is the real fix and needs `libc`; at one thread per timeout event, with timeouts now far rarer thanks to the 5-minute TTL, I'd rather not add that here. The helper's doc comment now says so instead of claiming a guarantee it doesn't provide.

## Nits

- Both missing tests added: closes-stdout-and-lingers, and 200KB of stdout past the pipe buffer.
- `agent_usage.rs` is now `rustfmt --check` clean. `build.rs` is still unformatted on main; leaving that alone.
- CI does not currently run these tests at all: the Rust job stops at `cargo check`. #46 adds `cargo test` and is open with checks green, so landing it before this branch is what makes any of the above run on a PR.

58 Rust tests pass, `tsc --noEmit` and `npm run build` clean.

PR #56, opened 2026-08-12
fix(claude): stop redeeming the Claude refresh token, which could log Claude Code out

### LINKED ISSUE #55: Claude quota refresh discards the rotated refresh token, which can log Claude Code out
## Context

The Claude quota provider refreshes the Claude OAuth access token when the stored credential has expired, but it never writes the refreshed credential back to the source it was read from. The code documents this as intentional:

```swift
/// Refreshed credentials are NOT persisted (mirror: the Rust code never
/// writes Claude credentials back anywhere).
static func refreshClaudeCredentials(
```

<sub>[`ClaudeProvider.swift:402-405`](https://github.com/handlecusion/tokcat/blob/v0.2.4/native/LocalPackage/Sources/Collector/Quota/ClaudeProvider.swift#L402-L405) (v0.2.4)</sub>

The refresh response can carry a **new** refresh token, and Tokcat does read it:

```swift
refreshToken: json["refresh_token"]?.asString ?? credentials.refreshToken,
```

<sub>[`ClaudeProvider.swift:441`](https://github.com/handlecusion/tokcat/blob/v0.2.4/native/LocalPackage/Sources/Collector/Quota/ClaudeProvider.swift#L441)</sub>

That value only lives in the returned `ClaudeCredentials`, which goes out of scope once `fetchInner()` returns. If the authorization server rotates refresh tokens — issuing a new one and retiring the one that was presented — then the credential still sitting in the Keychain is a token that has **already been redeemed**. The next refresh performed by Claude Code itself can then be rejected, and the user is forced to sign in again.

In other words, a passive usage monitor can consume the token lifecycle of the CLI it is monitoring.

> Scope: this concerns the Claude OAuth quota path only. Local token history parsing from session logs is unaffected.

## Observed Behavior

| Check | Result |
|---|---|
| Claude Code OAuth login | Drops repeatedly and needs re-authentication, without the user having logged out |
| Agent limits card | Intermittently shows `Claude OAuth refresh failed. Run 'claude' to re-authenticate.` |
| Observed access token lifetime | ~8 hours |
| Tokcat auto-refresh interval | 30 minutes ([`QuotaStore.swift:22`](https://github.com/handlecusion/tokcat/blob/v0.2.4/native/LocalPackage/Sources/Model/Stores/QuotaStore.swift#L22), [`DashboardStore.swift:50`](https://github.com/handlecusion/tokcat/blob/v0.2.4/native/LocalPackage/Sources/Model/Stores/DashboardStore.swift#L50)) |
| Expiry check | `now >= expiresAt`, no safety margin ([`ClaudeProvider.swift:396-400`](https://github.com/handlecusion/tokcat/blob/v0.2.4/native/LocalPackage/Sources/Collector/Quota/ClaudeProvider.swift#L396-L400)) |
| Net effect | When the CLI is idle, Tokcat is normally the **first** party to observe the expired token, so it is the party that redeems the refresh token |

The failures cluster after the machine has been idle longer than the access token lifetime — i.e. exactly when Claude Code has not been running to refresh the credential itself.

## Suspected Cause

```mermaid
sequenceDiagram
    participant T as Tokcat (polls every 30 min)
    participant S as Auth server
    participant K as Keychain (Claude Code-credentials)
    participant C as Claude Code CLI

    Note over K: access token expired,<br/>refresh token RT1 stored
    Note over C: not running — nothing refreshes the credential

    T->>K: read credential
    T->>S: POST /oauth/token (grant_type=refresh_token, RT1)
    S-->>T: new access token + RT2 (RT1 retired)
    Note over T: RT2 parsed at :441,<br/>then dropped — nothing is written back
    Note over K: still holds RT1 (already redeemed)

    C->>K: read credential (RT1)
    C->>S: refresh with RT1
    S-->>C: rejected
    Note over C: login cleared → user must sign in again
```

Trigger conditions:

1. Claude Code is not running for longer than the access token lifetime.
2. Tokcat keeps polling and reaches the expired credential first ([`ClaudeProvider.swift:203`](https://github.com/handlecusion/tokcat/blob/v0.2.4/native/LocalPackage/Sources/Collector/Quota/ClaudeProvider.swift#L203)).
3. Tokcat redeems the refresh token and discards what the server returned.
4. The stored credential is left pointing at a spent refresh token.

A secondary effect: because nothing is persisted, every subsequent 30-minute poll retries the refresh with the same stale token, which can also get the refresh endpoint to rate-limit.

## What is verified vs. inferred

- **Verified from the source:** the refreshed credential — including any rotated `refresh_token` — is never persisted, by design.
- **Inferred:** that the authorization server retires the presented refresh token on rotation. This is server-side behavior that cannot be confirmed from outside, but the fact that the response carries a `refresh_token` at all (and that Tokcat parses it) indicates rotation is in play, and the observed symptom matches.

## Note on the credential lookup order

[#16](https://github.com/handlecusion/tokcat/issues/16) correctly changed the loader to prefer the Keychain over `~/.claude/.credentials.json` ([`ClaudeProvider.swift:280-296`](https://github.com/handlecusion/tokcat/blob/v0.2.4/native/LocalPackage/Sources/Collector/Quota/ClaudeProvider.swift#L280-L296)). A side effect worth noting here: because the Keychain branch returns first, keeping the JSON file mirrored/fresh externally has no effect on this path — Tokcat always reads, and refreshes from, the authoritative credential.

## Expected Behavior

Reading usage should never change the authentication state of the tool being monitored. Tokcat should either leave the credential lifecycle entirely to Claude Code, or take full responsibility for it — the current middle ground consumes a rotation without recording it.

## Suggested Fixes

**Option A — do not refresh at all (preferred).** Treat an expired access token as "usage temporarily unavailable" and surface something like `Claude access token expired — open Claude Code to refresh`. Tokcat is a read-only monitor; the token lifecycle belongs to the CLI that owns the login. This requires no new permissions, removes the possibility of invalidating the user's session, and also eliminates the repeated-refresh rate-limiting.

**Option B — persist the rotated credential.** Write the refreshed credential back to the source it was read from (Keychain, or the file when that was the source). This keeps the quota card working while the CLI is idle, but it is the riskier option: it needs write access to the Keychain item, and it has to be concurrency-safe against Claude Code's own writes. A partial or racing write could break the very login it is trying to preserve.

**Option C — interim mitigation, regardless of A or B.** Add a setting to disable the Claude quota provider, so affected users can opt out of this path without quitting the whole app. Today the only workaround is to stop running Tokcat.

## How this could be confirmed

Temporarily log whether `json["refresh_token"]` in the refresh response differs from the refresh token that was submitted. If it does, rotation is confirmed and the currently discarded value is the credential the system should be holding.

## Environment

- Tokcat 0.2.x (source references above are against tag `v0.2.4`)
- macOS 13+ (Apple Silicon)
- Claude Code, OAuth/subscription login, credential stored in the macOS Keychain

oesni:
Fixes #55.

Tokcat could log Claude Code out. The Claude quota provider refreshed an expired access token but never persisted the result — documented in the code as intentional — while the refresh response can carry a **rotated** refresh token that was parsed into a local value and dropped. When the CLI is idle past the access token lifetime, Tokcat polls every 30 minutes and is normally the first party to reach the expired credential, so it redeemed the refresh token, retiring it server-side while the Keychain kept the spent one. The next refresh Claude Code performed could then be rejected, and the user was silently signed out.

Reading usage must not change the authentication state of the tool being monitored.

## Approach

The refresh path is **removed** rather than made to persist. Persisting was the other option, but the Claude credential lives in the login Keychain that Claude Code owns and writes under its own locking — racing that write risks breaking the very login it would be trying to preserve. Codex is deliberately left as it is: its credential is a plain file this app can safely rewrite, which is why `CodexProvider` still refreshes and `saveCodexCredentials` still persists.

The trade-off is that an expired token now surfaces as an error instead of being transparently renewed. That is the point — `claude` owns the token lifecycle, so the error tells the user to run it, and deliberately says *refresh*, not *re-authenticate*, because the login itself is still intact.

## Commits

**1. `fix(claude): stop redeeming the Claude refresh token`**
- Drops `refreshClaudeCredentials` and its endpoint/client-id constants.
- Drops `ClaudeCredentials.refreshToken` and stops lifting it out of the credential blob, so the token cannot be redeemed *by construction* rather than by convention.
- Adds `validateCredentials`, splitting the expiry and scope pre-flight out of `fetchInner` so both are testable without a network round-trip.

**2. `fix(quota): refresh quota on manual Refresh, tighten Claude error tests`**
- Telling the user to run `claude` was useless while no Refresh path could pick the result up: all three user-initiated entry points (header button, Refresh Now menu item, ⌘R) drove `DashboardStore` only, so quota recovery waited out the 5-minute panel staleness check or the 30-minute timer. They now drive `QuotaStore` too.
- `SettingsSheet`'s refresh stays dashboard-only — it is a targeted graph rebuild after the Cursor cache write, not a user Refresh.
- The header spinner tracks both stores, since quota is the slower one and is usually why Refresh was pressed.

**3. `fix(quota): queue an explicit Refresh instead of dropping it mid-fetch`**
- Wiring alone was still not enough: `QuotaStore`'s re-entry guard discards any call arriving while a fetch is in flight, and that fetch may have read its credentials before the user fixed them — so the one press that mattered could still be answered with the stale result.
- `refresh()` keeps dropping, which is what `start()`, the timer and `refreshIfStale()` want; they must not stack four upstream calls plus three subprocess spawns. The new `refreshNow()` records the press, and `refresh()` drains it with one more pass before finishing.
- Repeated presses collapse into a single extra pass, and `isRefreshing` stays true across the hand-off so the spinner does not blink.
- `runProviders` becomes an injectable seam — internal, so no new public API — because the store had no tests at all and the state machine is what carries the risk here.

## Testing

`swift test` in `native/LocalPackage`: **149 tests / 27 suites** pass (143 before).

- `QuotaClaudeTests`: expiry reported rather than renewed, expiry pinned ahead of the scope check, and both rejection messages asserted verbatim — that copy is the entire remedy the user gets, so a wording change should require an intentional test update.
- `QuotaStoreRefreshTests` (new): drop-while-busy (unchanged contract), queue on explicit press, coalescing, spinner continuity, and the idle path. The harness uses no wall-clock sleeps — it pumps the main-actor run loop, bounded, and fails explicitly on timeout rather than hanging.

Both new test groups were run against a reverted fix to confirm they are not vacuous: 3 of the 5 `QuotaStoreRefreshTests` fail without it, and the 2 that still pass are the ones covering pre-existing behavior.

> **Note on the app target:** `native/Tokcat/TokcatApp.swift` is outside the SwiftPM package and I did not have `xcodegen` available, so it was verified with `swiftc -typecheck -parse-as-library` against the built module search path (exit 0, with a negative control confirming the check catches errors) rather than a real `xcodebuild`. Worth a proper build on your side.

## Known cosmetic gap, not addressed

`HeaderBar`'s `.onChange` does not fire for the initial value, so if the header mounts while a quota refresh is already active the spinner is missing for that first pass. It cannot stick or flicker — only fail to appear. Left alone to keep this PR to the auth fix; happy to follow up if you'd rather have it here.

### DIFF
--- a/native/LocalPackage/Sources/Collector/Quota/ClaudeProvider.swift
+++ b/native/LocalPackage/Sources/Collector/Quota/ClaudeProvider.swift
@@ -3,15 +3,29 @@ import Foundation
 
 // Port of the Claude OAuth quota provider in agent_usage.rs
 // (:659-742, 786-863, 957-995, 1057-1100, 1192-1315, 2209-2236).
+//
+// One deliberate divergence from that port: the Rust code refreshed an expired
+// Claude access token, and this no longer does. See `ClaudeCredentials` (#55).
 
 private let claudeUsageURL = "https://api.anthropic.com/api/oauth/usage"
-private let claudeRefreshURL = "https://platform.claude.com/v1/oauth/token"
-private let claudeClientID = "9d1c250a-e61b-44d9-88ed-5944d1962f5e"
 private let claudeKeychainService = "Claude Code-credentials"
 
+/// Tokcat reads the Claude credential but never renews it.
+///
+/// Redeeming the refresh token rotates it server-side, and Tokcat has no way
+/// to write the replacement back: the credential lives in the login Keychain,
+/// which Claude Code owns and updates under its own locking. A refresh here
+/// would leave the stored credential pointing at a token that has already been
+/// spent, and the next refresh Claude Code performs would be rejected —
+/// silently logging the user out (#55).
+///
+/// So `refreshToken` is deliberately absent from this type. An expired access
+/// token is surfaced as an error telling the user to run `claude`, which is the
+/// process that owns the token lifecycle. Codex is different (its credential is
+/// a plain file this app can rewrite), which is why `CodexProvider` refreshes
+/// and `saveCodexCredentials` persists.
 struct ClaudeCredentials {
     var accessToken: String
-    var refreshToken: String?
     var expiresAt: Date?
     var scopes: [String]
     var rateLimitTier: String?
@@ -198,15 +212,8 @@ public enum ClaudeQuotaProvider {
     }
 
     static func fetchInner() async throws -> AgentUsageSnapshot {
-        var credentials = try loadClaudeCredentials()
-        if claudeCredentialsExpired(credentials) {
-            credentials = try await refreshClaudeCredentials(credentials)
-        }
-
-        if !credentials.scopes.isEmpty && !credentials.scopes.contains("user:profile") {
-            throw QuotaError(
-                "Claude OAuth token lacks the user:profile scope. Run `claude logout && claude login`.")
-        }
+        let credentials = try loadClaudeCredentials()
+        try validateCredentials(credentials)
 
         var request = URLRequest(url: URL(string: claudeUsageURL)!)
         request.timeoutInterval = 30
@@ -317,7 +324,7 @@ public enum ClaudeQuotaProvider {
             .map { $0.trimmingCharacters(in: .whitespaces) }
             .filter { !$0.isEmpty }
         return ClaudeCredentials(
-            accessToken: accessToken, refreshToken: nil, expiresAt: nil,
+            accessToken: accessToken, expiresAt: nil,
             scopes: scopes, rateLimitTier: nil, subscriptionType: nil)
     }
 
@@ -382,11 +389,10 @@ public enum ClaudeQuotaProvider {
         } else {
             scopes = []
         }
+        // `refreshToken` is intentionally not read out of the blob: nothing in
+        // this app may redeem it. See the note on `ClaudeCredentials`.
         return ClaudeCredentials(
             accessToken: accessToken,
-            refreshToken: oauth["refreshToken"]?.asString
-                .map { $0.trimmingCharacters(in: .whitespaces) }
-                .flatMap { $0.isEmpty ? nil : $0 },
             expiresAt: expiresAt,
             scopes: scopes,
             rateLimitTier: oauth["rateLimitTier"]?.asString,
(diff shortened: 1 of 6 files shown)


handlecusion:
Thank you — this is excellent work. 🙇

The diagnosis lands: Tokcat polling first meant it was the party that redeemed the refresh token, rotating it server-side while the Keychain kept the spent one. Removing the refresh path rather than persisting it is the right call, and the Codex distinction (a plain file this app can safely rewrite) holds up. Reading usage must not mutate the auth state of the tool being monitored.

Verified locally:
- `swift test`: 149 tests / 27 suites pass — matches the number in the description.
- The app target you flagged as unverified: `xcodegen generate && xcodebuild -scheme Tokcat` → **BUILD SUCCEEDED**. That open item is closed.
- No leftover `refreshToken` references; `HeaderBar` is only used inside `DashboardRootView`, which already injects `QuotaStore`, so the new `@EnvironmentObject` is safe; and the drain loop is entirely `@MainActor` with no `await` between `apply(payload)` and the `pendingUserRefresh` check, so a press landing mid-fetch cannot be lost.

Agreed on leaving the initial-mount spinner gap out of this PR — it can only fail to appear, never stick or flicker. The one thing I would like to follow up on separately is that an expired token now replaces the card contents with an error string (`fetch()` returns a snapshot with `windows: []`); keeping the last good data dimmed with an error badge would soften that. Not a blocker here.

Really thorough work — thanks again!

PR #67, opened 2026-08-26
feat(settings): hide unused agents from usage and limits

oesni:
Unused clients still show up in the Token Usage stack and as erroring Agent limits tiles, so the yearly totals and the quota row include agents the user is not actually using.

Usage and limits answer different questions, so they are independent. An agent whose quota endpoint keeps failing can still have real token history, and a quota-only agent has no history at all.

## Approach

- `hiddenAgents` is the master off switch. `agentScopes` narrows an agent that is still on (`usage_and_limits` / `usage_only` / `limits_only`) so the chosen surface survives the toggle.
- Only offer a surface the agent can reach. Log-only clients lock to Usage only; Cursor has no limits surface while Fetch from cursor.com is off.
- Configuration lives only in Settings › Agents: per-row toggle + scope chip. The cards report a read-only `· N hidden`.
- Anything that reads quota — tiles, plan radios, menubar plan percent — goes through the limits-side predicate. A stranded pin falls back to Auto, including after a refresh that re-honours a stored `usage_only`, without dropping a still-valid pin on restart.
- Settings rows sort by the last 7 days (today inclusive). Lifetime totals stay as a tie-break.

<p align="center">
  Settings › Agents: per-agent usage and limits visibility
</p>

## Commits

**1. `feat(settings): persist per-agent usage and limits visibility`**
- `AppSettings` + `DashboardStore` own the scopes, inventory, and plan-pin clamp.

**2. `feat(ui): hide unused agents from Settings`**
- Settings › Agents, filtered cards/tabs, README screenshot.

**3. `test: cover agent visibility scopes and plan-pin clamps`**
- 22 cases, including restart vs in-session pin consistency.

## Testing

`swift test` in `native/LocalPackage`: **200 tests / 35 suites** (22 new in `AgentVisibilityTests`).

## Not in this PR

- No `MARKETING_VERSION` / `CURRENT_PROJECT_VERSION` bump — that belongs in a release commit.
- The Live session card is unfiltered. It is a live tail, not the usage or limits surface.
- A SwiftUI `Menu` still will not open in the nonactivating tray panel; the new chip uses the existing `AppKitMenu` helper.


handlecusion:
Reviewed and pushed two follow-up commits.

## `fix:` review findings

**Year-blind usage tab** (`DashboardRootView.swift:155`). The Token Usage chart was gated on `!isAgentHiddenFromUsage`, which reads all-time totals, while `activeBars`/`stats` beside it are year-filtered. An agent whose logs live entirely in another year but which owns a tab through its limits tile drew an empty chart over a zeroed StreaksCard — a tab that did not exist before this PR. Now gated on `presentClients`, which already means "in this year and not hidden from usage".

**Untrimmed omp home overrides** (`OmpParser.swift:44,48`). `PI_CODING_AGENT_DIR`, `PI_CONFIG_DIR` and `TOKCAT_OMP_HOMES` were emptiness-checked with `rustTrim` but appended untrimmed, so one stray space from a shell export or a dotenv line built a path that never resolves and every omp session went missing with no error. An absolute `PI_CONFIG_DIR` was also concatenated onto `$HOME` rather than used as the root.

Everything else held up — I traced `apply` → `planPinIsStranded` → `settings.didSet` → `recomputeDerived` reentry and found no non-terminating loop, no stale derived state, and no path that skips the tray-title recompute.

## `feat(menubar):` the gap the PR left open

`computeTrayTitle` takes `visibleAgentUsage` for the plan percentage but still took the raw `liveTokensPerMin` for the rate, so in `tokens_per_min` mode a hidden agent kept moving the menubar. The number is a usage reading, so it now follows the usage switch.

`DashboardStore.clientsHiddenFromUsage` is the exclusion set — deliberately not `agentsHiddenFromUsage`, which is the card's display count and lists only agents with history to hide, while an agent with no logs at all can be burning tokens right now. It is seeded in `init` so a hidden agent gets no free window between launch and the first collector pass.

The trace itself stays unfiltered. The Live activity card is a tail, not the usage surface — that is the call this PR already made, and it holds.

### The tick pays nothing for it

`ratePerMin` + `rateInWindow` + `trace` answered three questions from one event ring in three walks. `liveSnapshot` answers all three in one, carrying the 60s subtotal inside the trace group and folding the per-client rollups out of the groups afterwards, so hashing stays at one lookup per event. Reading a single `now()` also makes the three values mutually consistent, which three separate actor hops did not guarantee.

Measured on a full 1h ring (14,399 events, release build, 200 iterations):

| | ms per tick |
|---|---|
| before (three walks) | 0.216 |
| after (`liveSnapshot`) | 0.192 |

The first attempt kept separate per-client dictionaries and came out *slower* (0.321ms) — three hash inserts per event beat two cheap scans. The single-dictionary version is the one that landed.

## Testing

`swift test` in `native/LocalPackage`: **215 tests / 38 suites**, 15 new. They cover `liveSnapshot` agreeing with the three calls it replaces down to the `Float` round-trip the tray title formats, the 60s window staying narrower than the 600s one, hidden ids being matched only after `TraceClients.normalize` (so `claude-code` does not hide `claude` by accident), a toggle republishing without waiting out a 5s tick, a scope that inventory clamps away not leaking into the live filter, and the set surviving a restart before any refresh.

`xcodebuild -scheme Tokcat` also passes.

PR #79, opened 2026-08-31
harness: force review-prompt fetch refspecs

handlecusion:
One-liner from the auto-review note on #78: non-forced refspecs can keep a stale `pr-<N>` ref after a force-push, producing a silently wrong diff. Both fetches now use `+`.

### DIFF
--- a/.claude/harness/REVIEW_PROMPT.md
+++ b/.claude/harness/REVIEW_PROMPT.md
@@ -25,12 +25,12 @@ Therefore:
 
 ```sh
 # Pin main to a named ref — a later fetch would overwrite FETCH_HEAD.
-git fetch origin main:refs/remotes/origin/main
+git fetch origin +main:refs/remotes/origin/main
 git show refs/remotes/origin/main:.claude/harness/REVIEW_PROMPT.md   # canonical instructions
 git show refs/remotes/origin/main:AGENTS.md                          # canonical invariants
 git diff refs/remotes/origin/main...HEAD                             # the diff you are reviewing
 # fork PR whose head is not checked out:
-git fetch origin "pull/<N>/head:refs/heads/pr-<N>"
+git fetch origin "+pull/<N>/head:refs/heads/pr-<N>"
 git diff refs/remotes/origin/main...pr-<N>
 ```
 


.claude/harness/REVIEW_PROMPT.md
```diff
@@ -25,12 +25,12 @@ Therefore:
...
 ```sh
 # Pin main to a named ref — a later fetch would overwrite FETCH_HEAD.
-git fetch origin main:refs/remotes/origin/main
+git fetch origin +main:refs/remotes/origin/main
 git show refs/remotes/origin/main:.claude/harness/REVIEW_PROMPT.md   # canonical instructions
 git show refs/remotes/origin/main:AGENTS.md                          # canonical invariants
 git diff refs/remotes/origin/main...HEAD                             # the diff you are reviewing
 # fork PR whose head is not checked out:
-git fetch origin "pull/<N>/head:refs/heads/pr-<N>"
+git fetch origin "+pull/<N>/head:refs/heads/pr-<N>"
```
handlecusion, line 33: `[q]` The `+` is right and load-bearing here, but it doesn't cover one residual case: force does **not** override git's refusal to fetch into a **checked-out** branch. Verified on git 2.43:

```
$ git checkout pr-79
$ git fetch origin "+pull/79/head:refs/heads/pr-79"
fatal: refusing to fetch into branch 'refs/heads/pr-79' checked out at '...'
# exit=128
```

That's a hard failure rather than a stale ref, so it's much less dangerous than the bug this PR fixes — and in practice the harness checks out the PR's own head ref (`harness/review-refspec`), never `pr-<N>`, so it likely never fires.

If you want the whole class gone, land the destination outside `refs/heads/*`, which can't be "checked out" and doesn't litter the branch namespace:

```sh
git fetch origin "+pull/<N>/head:refs/remotes/pr/<N>"
git diff refs/remotes/origin/main...refs/remotes/pr/<N>
```

Same exit-0 result in my test. Note `+` is still required even there — see the review body; `refs/remotes/*` does **not** get the force exemption the git-fetch docs imply. Entirely optional; the PR as written is correct.

---

handlecusion (commented):
🤖 **Claude Code** — 자동 리뷰

This makes both fetch refspecs in the review prompt forced. I reproduced the bug on git 2.43 and **both `+` are load-bearing** — including the one on `main:refs/remotes/origin/main`, which I expected to be a no-op: the git-fetch docs state that "any updates outside of `refs/{tags,heads}/*` will be accepted without `+`", but in practice a non-fast-forward update to `refs/remotes/origin/main` is rejected exactly like the `refs/heads/pr-<N>` one. So the second hunk isn't just symmetry — it fixes the same latent bug on the path *every* review takes.

One refinement to the PR description: the failure isn't silent at the shell level — git prints `! [rejected] ... (non-fast-forward)` and exits 1. It's the *consequence* that's silent, and that's the part that matters: the ref keeps its old value, and the `git diff ...` on the next line runs happily against the stale ref and yields a plausible-but-wrong diff. Accurate framing, worth stating precisely in a file whose whole job is telling a reviewer how to get a trustworthy diff.

Checked for the same pattern elsewhere on `main`: `ROUTINE_PROMPT.md:37` already uses `+refs/heads/$B:refs/remotes/origin/$B`, and its line 40 `git fetch origin main` relies on the configured (already-forced) refspec — so this brings the two harness docs in line rather than leaving a straggler. No AGENTS.md invariants in scope (no Swift, no version bump, no release assets), no test coverage applicable, no README/llms.txt drift.

Nothing blocks merging. One optional `[q]` inline about a residual edge that `+` alone doesn't cover.

<details>
<summary>Verification (git 2.43, synthetic remote force-moved to unrelated history)</summary>

```
=== NON-FORCED ===
  main->remotes/origin/main   git exit=1   ! [rejected] (non-fast-forward)
  pull/79->heads/pr-79        git exit=1   ! [rejected] (non-fast-forward)
  refs after: origin/main=2ceb72c pr-79=2ceb72c   (both STALE)
=== FORCED (+) ===
  main  git exit=0
  pr-79 git exit=0
  refs after: origin/main=b833053 pr-79=b833053   (both fresh)
```
</details>



---

PR #75, opened 2026-08-31
README: clarify merge gate wording

handlecusion:
Small wording tweak in the Delegating issues to Claude section.

PR #80, opened 2026-08-31
harness: document which PR webhook events actually deliver

handlecusion:
Empirical note from trying to retro-review #74: `pull_request.opened` delivered 5/5; `ready_for_review` (3 attempts) and `reopened` (1 attempt, dedicated trigger) never created a session. Documented in the harness README so nobody chases this again.

### DIFF
--- a/.claude/harness/README.md
+++ b/.claude/harness/README.md
@@ -63,7 +63,11 @@
    (`add_issue_comment`, `issue_write`, `create_pull_request`, `pull_request_review_write`, …)로 쓴다.
 
 5. **루틴 (PR 자동 리뷰)**: https://claude.ai/code/routines/trig_01BswRvckXcbV6xSM9CkLQxQ — "tokcat · PR auto-review".
-   GitHub 트리거 `pull_request.opened` / `ready_for_review` (Claude GitHub App 웹훅, Actions 불필요).
+   GitHub 트리거 `pull_request.opened` (Claude GitHub App 웹훅, Actions 불필요).
+   **실측 제약(2026-08-31)**: 구독을 걸어도 `ready_for_review`·`reopened`는 세션을 만들지 않았고
+   `opened`만 전달됐다(#75–#79에서 5/5). 즉 **웹훅 생성 이전에 열린 PR과 draft→ready 전환 PR은
+   자동 리뷰가 붙지 않는다** — 필요하면 리뷰 세션 링크 없이 사람이 보거나, 새 PR로 다시 연다.
+   또 웹훅 발화에는 시간당 상한이 있어(프리뷰) 몰리면 조용히 드롭된다.
    PR이 열리면 세션이 diff를 읽고 **COMMENT 리뷰**(인라인 + 총평, 마커 `kind=REVIEW`)를 남긴다.
    승인/변경요청은 절대 하지 않는다(머지 게이트와 분리). 프리뷰 기간엔 웹훅 발화에 시간당 상한이 있다.
    **트러스트 경계**: 같은 저장소 PR은 워크스페이스가 PR 헤드로 체크아웃된 채 시작되므로, 리뷰 세션은


.claude/harness/README.md
```diff
@@ -63,7 +63,11 @@
    (`add_issue_comment`, `issue_write`, `create_pull_request`, `pull_request_review_write`, …)로 쓴다.
 
 5. **루틴 (PR 자동 리뷰)**: https://claude.ai/code/routines/trig_01BswRvckXcbV6xSM9CkLQxQ — "tokcat · PR auto-review".
-   GitHub 트리거 `pull_request.opened` / `ready_for_review` (Claude GitHub App 웹훅, Actions 불필요).
+   GitHub 트리거 `pull_request.opened` (Claude GitHub App 웹훅, Actions 불필요).
```
handlecusion, line 66: `[invariant]` This line is the record of **what the routine is subscribed to**, and the diff rewrites it into a record of **what was observed to arrive**. Those are two different facts, and collapsing them loses the one a maintainer needs.

Concretely: if the routine's trigger config still lists `ready_for_review` (the PR body says the subscription was made, not that it was removed), then a future reader of this line sees `pull_request.opened` only, concludes the subscription is missing, and re-adds a subscription that is already there — the exact loop this PR is trying to end. The empirical note two lines down is worded as `구독을 걸어도 …` ("even with the subscription"), which confirms the subscription exists; line 66 now contradicts it.

Suggested fix — keep both facts on the line:

```
   GitHub 트리거 구독: `pull_request.opened` / `ready_for_review` — 실제 전달되는 건 `opened`뿐(아래 실측 제약).
```

If the `ready_for_review` subscription *was* actually deleted as part of this investigation, then line 66 is right as written and it's the `구독을 걸어도` phrasing on line 67 that should say so instead (past tense: 걸어 봤지만).

---

.claude/harness/README.md
```diff
@@ -63,7 +63,11 @@
...
+   **실측 제약(2026-08-31)**: 구독을 걸어도 `ready_for_review`·`reopened`는 세션을 만들지 않았고
+   `opened`만 전달됐다(#75–#79에서 5/5). 즉 **웹훅 생성 이전에 열린 PR과 draft→ready 전환 PR은
+   자동 리뷰가 붙지 않는다** — 필요하면 리뷰 세션 링크 없이 사람이 보거나, 새 PR로 다시 연다.
```
handlecusion, line 69: `[bug]` The finding is recorded here but the file that actually *acts* on it is left stale. `REVIEW_PROMPT.md:4` on `main` still opens with:

> You are a Claude Code cloud session fired by a GitHub trigger on `handlecusion/tokcat`: a pull request was opened (or marked ready for review).

After this PR the repo asserts both "`ready_for_review` never creates a session" (here) and "you may have been fired by `ready_for_review`" (the canonical prompt), 60 lines apart in the same directory. Since `REVIEW_PROMPT.md` is the file every review session reads first, it's the more load-bearing of the two.

Fix is one line in the same PR:

```
`handlecusion/tokcat`: a pull request was opened.
```

Two follow-ons worth doing in the same pass, because they're the same fact:

1. **The stored routine snapshot.** This README says at line 50–51 that when the prompt file changes, the routine's stored snapshot must be updated to match (웹 편집 또는 CLI `/schedule update`). The stored bootstrap for the auto-review routine currently carries the same "opened or marked ready for review" wording, so editing only the file leaves the snapshot as the third copy of the stale claim — and the snapshot is what runs when the clone lacks the file.
2. **`AGENTS.md`** is fine as-is — "Every opened PR also gets an automatic COMMENT review" happens to be accurate under the new finding. No change needed, noting it so it doesn't get swept up.

---

.claude/harness/README.md
```diff
@@ -63,7 +63,11 @@
...
+   또 웹훅 발화에는 시간당 상한이 있어(프리뷰) 몰리면 조용히 드롭된다.
```
handlecusion, line 70: `[q]` This sentence introduces a second mechanism that can produce the *exact* symptom the paragraph above attributes to event type — "몰리면 조용히 드롭된다". As written, a reader can't tell whether the 3 `ready_for_review` attempts failed because the event type isn't delivered, or because they landed in a saturated hour. That matters, because the two have opposite remedies (give up on draft→ready vs. just retry later).

The repo actually contains the evidence that resolves it, and it favors your conclusion — worth putting in the note so it survives:

- #75's auto-review was submitted `2026-08-31T07:19:58Z`, #79's `2026-08-31T07:38:41Z`.
- So the five `opened` firings all landed inside roughly one 20-minute window and none were dropped — the cap demonstrably was **not** exhausted in the window where the `ready_for_review` attempts were being made.

That makes "the event type isn't delivered" the well-supported reading rather than an inference the reader has to take on trust. Suggest folding it in:

```
`opened`만 전달됐다(#75–#79에서 5/5, 07:19–07:38Z 한 창(window) 안에서 — 즉 시간당 상한 소진이 원인이 아니다).
```

If the `ready_for_review` attempts were actually made *outside* that window, then this reasoning doesn't apply and the conclusion rests on the 3 attempts alone — in which case 아직 관측되지 않았다 is the more defensible wording than 세션을 만들지 않았다.

Separately, and much smaller: this line's rate-limit clause now says the same thing as line 72's `프리뷰 기간엔 웹훅 발화에 시간당 상한이 있다`, two lines apart. Worth dropping one.

---

handlecusion (commented):
🤖 **Claude Code** — 자동 리뷰

Docs-only change (+5/−1) recording that only `pull_request.opened` actually delivers a webhook to the auto-review routine. The finding itself checks out — I confirmed the 5/5 endpoints independently: #75's auto-review landed `07:19:58Z` and #79's `07:38:41Z`, so five `opened` firings inside one ~20-minute window all delivered.

Two things to look hardest at. First, the note is written into the README but `REVIEW_PROMPT.md:4` still tells every review session it may have been fired by `ready_for_review` — after this PR the harness directory asserts both halves of the contradiction, and the prompt file is the one that actually runs. Second, line 66 replaces a record of *what the routine subscribes to* with a record of *what arrives*; if the `ready_for_review` subscription is still configured (line 67's `구독을 걸어도` implies it is), that invites someone to re-add a subscription that already exists.

Neither blocks merging — this is documentation, no code, CI, or AGENTS.md invariants in scope (no Swift, no version bump, no release assets), and it leaves the repo better informed than it found it. But the `REVIEW_PROMPT.md` one-liner is cheap and belongs in this PR rather than a follow-up, along with the routine's stored snapshot that this README's own line 50–51 requires be kept in lockstep.

Reviewed by reading (Linux VM, no Swift toolchain) — not a limitation here, as nothing in the diff is compiled.



---

PR #81, opened 2026-08-31
harness: align review-prompt preamble with measured webhook delivery

handlecusion:
Consistency nit from the auto-review on #80: the prompt still claimed ready_for_review as a firing event.

PR #82, opened 2026-08-31
harness: no-finding reviews are one line, no process narration

handlecusion:
Owner request after #81's review: when there is nothing to flag, the review is exactly one line and zero inline comments; with findings, findings only — no diff-restating or self-narration.

### DIFF
--- a/.claude/harness/REVIEW_PROMPT.md
+++ b/.claude/harness/REVIEW_PROMPT.md
@@ -68,10 +68,16 @@ changes-requested review would trip the harness's approval-withdraw logic).
 
 - Inline comments on the exact lines for each finding: what breaks, why, and
   a concrete fix. Severity-tag each (`[bug]`, `[invariant]`, `[test]`, `[q]`).
-- Review body: start with `🤖 **Claude Code** — 자동 리뷰`, then a 2–5 line
-  verdict (what this PR does, the one or two things to look hardest at, and
-  whether anything blocks merging). If you found nothing: say so in two lines
-  and post no inline comments.
+  An inline comment must carry an actionable finding — no observations,
+  reminders, or "follow-ups" without a concrete change to make.
+- Review body: start with `🤖 **Claude Code** — 자동 리뷰`, then the verdict.
+  **Findings only.** Never restate what the PR does, never narrate your own
+  process ("verified against …", "I checked …", "worth noting …"), never
+  praise, never pad.
+  - With findings: 2–5 lines — the one or two things to look hardest at and
+    whether anything blocks merging.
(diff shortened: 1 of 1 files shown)


.claude/harness/REVIEW_PROMPT.md
```diff
@@ -68,10 +68,16 @@ changes-requested review would trip the harness's approval-withdraw logic).
...
+  An inline comment must carry an actionable finding — no observations,
+  reminders, or "follow-ups" without a concrete change to make.
+- Review body: start with `🤖 **Claude Code** — 자동 리뷰`, then the verdict.
+  **Findings only.** Never restate what the PR does, never narrate your own
+  process ("verified against …", "I checked …", "worth noting …"), never
+  praise, never pad.
+  - With findings: 2–5 lines — the one or two things to look hardest at and
+    whether anything blocks merging.
+  - **No findings: exactly one line — `특이사항 없음.` / `Nothing to flag.` —
+    and zero inline comments.** Resist the urge to prove you looked.
```
handlecusion, line 80: `[bug]` **"exactly one line" collides with the marker rule two bullets down (L81–82).**

Bolded, with `exactly` and a matching `zero inline comments`, this reads as a cap on the whole body — not just the verdict. A session following it literally posts `Nothing to flag.` and drops `<!-- claude-harness kind=REVIEW pr=<N> -->`. The marker is load-bearing in two places:

- **This same file's dedupe rule** — "If a Claude review with that marker already exists on the PR, stop instead of duplicating it." A no-finding review without the marker is invisible to that check, so a re-fire (reopened PR, retried webhook) posts a duplicate review. No-finding reviews are exactly the ones this branch produces, so the gap lands where it's most likely to be hit.
- **`.github/workflows/claude-followup.yml:78`** — `grep -qE '^<!-- claude-harness'` skips marker-carrying text as a harness post rather than a human request. Secondary here, since the `@claude` check gates PR reviews first, but it's the intended second layer.

The `With findings` branch doesn't have this problem: "2–5 lines" reads as a length target, not a prohibition on the trailing marker.

Concrete fix — scope the count to the verdict and name the exemption:

```md
  - **No findings: one line of verdict — `특이사항 없음.` / `Nothing to flag.` —
    plus the marker line, and zero inline comments.** Resist the urge to prove you looked.
```

---

handlecusion (commented):
🤖 **Claude Code** — 자동 리뷰

Look hardest at the no-finding branch: the bolded "exactly one line" reads as a cap on the whole body, which drops the `claude-harness` marker that this same file's dedupe rule depends on — and no-finding reviews are precisely the ones that branch produces. Wording fix inline.

Nothing blocks merging.



---

PR #84, opened 2026-09-05
Track Aside browser usage

handlecusion:
## Track Aside browser usage

Aside (the AI browser) keeps a JSONL transcript per browser-agent session under `~/.aside/u/<account>/sessions/<date>_<id>/messages.jsonl`, and every `role == "assistant"` record carries the token usage **and the price** of exactly one API response. That is the same per-call shape `omp` writes, minus the `type`/`message` envelope — so it drops straight into the collector as a twelfth client, with no network access and no new dependency.

### What's in here

- `Collector/Parsers/AsideParser.swift` — new `aside` client. Roots are `~/.aside` plus `TOKCAT_ASIDE_HOMES`, one sessions root per account dir (`u/0`, `u/1`, …), deliberately rooted at each account's `sessions/` so the walk never touches the sibling `memory/` and `passwords/` trees. Dedup key is `aside:<responseId>`.
- `UsageGraph.registry` — registered after `omp`; both stay out of the Rust parity rosters, since neither has a `src-tauri` counterpart.
- `Tail/UsageTailer.swift` — `.aside` tail kind and `parseAsideLine`, real + `simulatedHome` roots, and a `messages.jsonl` name filter so a stray `artifacts/*.jsonl` download can never be tailed as a transcript.
- `UserInterface/Theme/ClientStyle.swift` — `"aside": ("Aside", "#14b8a6", nil)`.
- `Tests/CollectorTests/AsideTests.swift` — 10 tests (parse, cost passthrough, gateway fallthrough, zero-token skip, provider mapping, per-account roots, tailer + retry dedup).
- README / landing page / `llms.txt` + `llms-full.txt` client lists: 11 → 12 clients.

### Two decisions worth reviewing

- **`usage.reasoning` is dropped.** `totalTokens` is exactly `input+output+cacheRead+cacheWrite`, which makes `reasoning` a *subset* of `output` — 3,202 of my 7,672 records carry one, and passing it through would bill those tokens a second time at the output rate in `estimateCost`. Same call `OmpParser` makes.
- **`provider` is a routing account, not a vendor.** `"claude-code"` maps to `anthropic`; Aside's own gateway maps to `""` so the model string decides. Gateway calls are plan-covered and record `cost.total == 0`, which falls through to the bundled price table exactly like any other subscription client.

### Screenshots

The Aside tab, with its own totals, chart and streaks:

Aside client tab

The live tailer picking Aside up alongside the CLIs (both rows are real, from one tick):

Live session card with Aside rows

### Verification

- `tokcat-dump graph --clients aside` reconciles **per-day, exactly** against an independent pass over the same 763 transcripts. The only deltas are intended: 54 zero-token aborted turns dropped, 1 duplicate `responseId` deduped, and a $0.15 bundled-price estimate across the 82 zero-cost gateway calls.
- `tokcat-dump tail-sim` on a real session fixture: 6 events, 134,048 tokens, `client=aside`, `agent=main`, model date-stamp normalized — byte-for-byte that session's own six calls.
- No double counting with the Claude Code parser: Aside's `responseId`s appear nowhere under `~/.claude`, because it calls the API itself with the subscription token.
- `swift build` + full suite green (263 tests, 10 new); app target built via `xcodegen` + `xcodebuild` and driven by hand for the screenshots above. Rebased onto `main` after the 0.3.1 release and the Cursor live-session work; the only overlap was the tailer's source-list comment.

On my machine this surfaces **$367 / 1.23 B tokens over 27 active days** that Tokcat could not see before — the third-largest client after Claude Code and Codex CLI.


native/LocalPackage/Sources/Collector/Parsers/AsideParser.swift
```diff
@@ -0,0 +1,144 @@
...
+enum AsideParser: UsageParser {
+    static let clientName = "aside"
+
+    static func parse(_ cache: UsageCache?) -> [UsageMessage] {
+        var files: [String] = []
+        for root in asideSessionRoots() {
+            // Only `messages.jsonl` carries usage, but the predicate stays on
+            // the extension: the walk is already recursive and the sibling
+            // `artifacts/` payloads never end in .jsonl.
+            files.append(contentsOf: collectFiles(root) { rustExtension($0) == "jsonl" })
```
handlecusion, line 37: `[bug]` The parser walks `artifacts/`; the tailer refuses to. Only one of them can be right.

`collectFiles` recurses into every subdirectory (`Support/FileScan.swift:45-46`) and this predicate tests only the extension, so `<session>/artifacts/*.jsonl` is collected and parsed into the usage history.

The comment above justifies that with "the sibling `artifacts/` payloads never end in .jsonl" — but the rest of the PR assumes the opposite. `UsageTailer` adds an explicit `entry.name != "messages.jsonl"` guard for exactly this, and `AsideTests.swift:228` builds an `artifacts: ["notes.jsonl": …]` fixture to prove the guard works. Aside is a *browser* agent: a downloaded `.jsonl` is precisely the kind of thing that lands in `artifacts/`, and if it contains any `role == "assistant"` record with a `usage` object it becomes phantom spend in the graph. Dedup won't catch it either — a foreign transcript carries its own `responseId`s, so `aside:<id>` just makes it look like a distinct call. That also means the live tab and the historical graph can disagree on the same session.

Fix is the idiom `GrokParser.swift:21` already uses for the same `updates.jsonl`-among-siblings problem:

```swift
files.append(contentsOf: collectFiles(root) { rustFileName($0) == "messages.jsonl" })
```

Worth a parser-side test too — `writeAsideSession` already takes an `artifacts:` argument, but only the tailer suite passes it, which is why this gap is green today.

---

handlecusion (commented):
🤖 **Claude Code** — 자동 리뷰

`AsideParser.parse` filters on the `.jsonl` extension alone, so it collects each session's `artifacts/` directory — the one the tailer explicitly skips and has a test for. A downloaded `.jsonl` there lands in the history as phantom spend, and dedup won't stop it. One-line fix inline.

Separately, the branch conflicts with `main` (`README.md`, `docs/index.html`, and the `UsageTailer.swift` roots header). Main picked up the Cursor opt-in wording after this branched, so resolving toward this branch's side would silently drop "plus opt-in Cursor usage from cursor.com" and the matching privacy caveat from the README intro and the FAQ answers. Worth resolving by hand rather than taking a side wholesale.

Everything else — dropping `usage.reasoning`, the zero-cost gateway fallthrough, per-account roots — traces correctly by reading; the cost/pricing claims need macOS CI to confirm.



---

PR #87, opened 2026-09-05
Spin the refresh icon about the glyph's arc center

### LINKED ISSUE #86: Refresh button spin wobbles: arrow.clockwise rotates about its layout-box center, 0.86 pt above the arc center
## Symptom

The header Refresh button's spin doesn't look like a spin — the glyph visibly **orbits**: it
wobbles/drifts instead of turning in place. Most noticeable during a slow refresh (quota fetch),
which is when the spinner runs longest.

## Where

```swift
if let startedAt = spinStartedAt {
    TimelineView(.animation) { context in
        Image(systemName: "arrow.clockwise")
            .font(.system(size: 12, weight: .semibold))
            .rotationEffect(.degrees(
                context.date.timeIntervalSince(startedAt)
                    / 0.8 * 360))
    }
```

<sub>[`HeaderBar.swift:78-85`](https://github.com/handlecusion/tokcat/blob/v0.3.2/native/LocalPackage/Sources/UserInterface/Views/HeaderBar.swift#L78-L85) (v0.3.2)</sub>

## Why

`.rotationEffect(_)` defaults to `anchor: .center`, i.e. the center of the **layout box**. For
`arrow.clockwise` that is not the center of the arc the eye tracks: the arrowhead overshoots the
ring radius at the top, so the ink bounding box (and therefore the layout box) is taller above the
arc than below it. The pivot ends up **above** the arc's true center, and the glyph sweeps a small
circle once per revolution.

Measured by rendering the exact view with `ImageRenderer` and least-squares fitting a circle to the
ring pixels (arrowhead trimmed as outliers):

| pointSize | layout box | arc center | pivot error | wobble diameter |
| --- | --- | --- | --- | --- |
| 11 | 11.75 × 13.88 | (5.87, 7.68) | +0.74 pt low | 1.48 pt (13% of glyph width) |
| **12 (ours)** | 12.62 × 14.88 | (6.29, 8.30) | **+0.86 pt low** | **1.72 pt (14%)** |
| 18 | 19.12 × 22.50 | (9.59, 12.58) | +1.33 pt low | 2.66 pt (14%) |
| 24 | 25.38 × 30.00 | (12.73, 16.73) | +1.73 pt low | 3.46 pt (14%) |

The error is horizontally negligible (|dx| < 0.05 pt) and purely vertical, and it is **scale
invariant**: the arc center sits at `UnitPoint(x: 0.500, y: 0.558 ± 0.003)` of the layout box at
every size tested. So a single `UnitPoint` constant fixes it at any font size — only an SF Symbols
redesign of the glyph could move it.

Ruled out as causes:
- No layout jump on spin start/stop — the `TimelineView` branch and the plain `Image` branch
  measure identically (12.625 × 14.875 both).
- Not a rendering/interpolation artifact — the offset reproduces offscreen in `ImageRenderer`.

## Repro (headless, no app launch)

```swift
// swiftc -O probe.swift -o probe && ./probe
import SwiftUI
import AppKit

@MainActor func pixels(_ view: some View, scale: CGFloat) -> (CGSize, [(Double, Double)])? {
    let r = ImageRenderer(content: view); r.scale = scale
    guard let cg = r.cgImage, let ns = r.nsImage else { return nil }
    let rep = NSBitmapImageRep(cgImage: cg)
    var pts: [(Double, Double)] = []
    for y in 0..<cg.height where true {
        for x in 0..<cg.width {
            if let c = rep.colorAt(x: x, y: y), c.alphaComponent > 0.5 {
                pts.append(((Double(x) + 0.5) / scale, (Double(y) + 0.5) / scale))
            }
        }
    }
    return (ns.size, pts)
}

func fitCircle(_ p: [(Double, Double)]) -> (cx: Double, cy: Double, r: Double) {
    var sx = 0.0, sy = 0.0, sxx = 0.0, syy = 0.0, sxy = 0.0, sxz = 0.0, syz = 0.0, sz = 0.0
    let n = Double(p.count)
    for (x, y) in p {
        let z = x * x + y * y
        sx += x; sy += y; sxx += x * x; syy += y * y; sxy += x * y
        sxz += x * z; syz += y * z; sz += z
    }
    let a = [[sxx, sxy, sx], [sxy, syy, sy], [sx, sy, n]], b = [sxz, syz, sz]
    func det3(_ m: [[Double]]) -> Double {
        m[0][0] * (m[1][1] * m[2][2] - m[1][2] * m[2][1])
            - m[0][1] * (m[1][0] * m[2][2] - m[1][2] * m[2][0])
            + m[0][2] * (m[1][0] * m[2][1] - m[1][1] * m[2][0])
    }
    let d = det3(a)
    func sub(_ col: Int) -> [[Double]] { var m = a; for i in 0..<3 { m[i][col] = b[i] }; return m }
    let cx = det3(sub(0)) / d / 2, cy = det3(sub(1)) / d / 2
    return (cx, cy, (det3(sub(2)) / d + cx * cx + cy * cy).squareRoot())
}

MainActor.assumeIsolated {
    let view = Image(systemName: "arrow.clockwise").font(.system(size: 12, weight: .semibold))
    guard let (size, all) = pixels(view, scale: 8) else { fatalError("render failed") }
    var p = all, c = fitCircle(p)
    for _ in 0..<6 {                     // trim the arrowhead: keep the 60% closest to the ring
        let res = p.map { abs((($0.0 - c.cx) * ($0.0 - c.cx)
            + ($0.1 - c.cy) * ($0.1 - c.cy)).squareRoot() - c.r) }
        let cut = res.sorted()[Int(Double(res.count) * 0.6)]
        let keep = zip(p, res).filter { $0.1 <= cut }.map { $0.0 }
        if keep.count < 20 { break }
        p = keep; c = fitCircle(p)
    }
    print(String(format: "layout %.2f x %.2f | arc center (%.2f, %.2f) | anchor (%.4f, %.4f) | dy %+.3f pt",
                 size.width, size.height, c.cx, c.cy,
                 c.cx / Double(size.width), c.cy / Double(size.height),
                 c.cy - Double(size.height) / 2))
}
```

Prints `dy +0.860 pt` today; a correct pivot is `|dy| <= 0.05 pt` from whatever anchor the fix uses.

## Expected

The glyph turns about its own arc center — no positional drift at any angle.

## Constraints

- **macOS 13 is the floor** (`AGENTS.md`), so `.symbolEffect(.rotate)` (macOS 15+) is not an option.
- Keep the existing time-driven `TimelineView` spin. It exists specifically because a
  state-animated `repeatForever` stuttered on spin-up and unwound backward on stop — see the
  comment above the call site. Don't reintroduce that.
- The resting (non-spinning) icon must not move: at 0°/360° any anchor is the identity transform,
  so an anchor-only fix is position-preserving by construction. Verify this holds for whatever
  approach is chosen, and keep the refresh and gear buttons optically aligned with each other.
- `native/LocalPackage/Sources/UserInterface` is views-only — no logic layer changes needed.

## Verification asked for in the PR

1. The probe above, before and after, showing the pivot error collapse to ~0.
2. A visual check of the running app (`cd native && xcodegen generate && open Tokcat.xcodeproj`, or
   an equivalent build) with a real refresh in flight — the spin must read as in-place rotation.
3. `swift build && swift test` in `native/LocalPackage` (should be unaffected; there is no
   `UserInterfaceTests` target and this issue does not ask for one).

handlecusion:
🤖 **Claude Code**

## Summary

`.rotationEffect(_)` pivots about `anchor: .center` — the center of the **layout box**. `arrow.clockwise`'s arrowhead overshoots the ring at the top, so the ink box is taller above the arc than below it, and the default pivot lands 0.86 pt above the circle the eye tracks at our 12 pt. The glyph orbited a 1.72 pt circle (14 % of its width) once per revolution instead of spinning in place, most visible during a slow quota refresh.

The fix is one `anchor:` argument pinning the pivot to the measured arc center, `UnitPoint(x: 0.5, y: 0.558)`. Nothing else changes: the time-driven `TimelineView` spin stays exactly as it is (the state-animated `repeatForever` that stuttered on spin-up and unwound backward is not reintroduced), and at rest the rotation is the identity transform at any anchor, so the icon does not move and stays optically aligned with the gear.

## Changes

- `native/LocalPackage/Sources/UserInterface/Views/HeaderBar.swift`
  - New `private static let refreshArcCenter = UnitPoint(x: 0.5, y: 0.558)`, documented with the measurement that produced it and the note that only an SF Symbols redesign of the glyph should move it.
  - The spinning branch's `.rotationEffect` now passes `anchor: Self.refreshArcCenter`.

This is the only rotated glyph in `native/` — `grep -rn "rotationEffect" native/` returns just this call site — so there is nothing else to correct.

## Assumptions

- **The constant is tuned for the 12 pt we draw, not averaged across sizes.** The issue's table gives arc-center ratios of 0.5535 / 0.5578 / 0.5591 / 0.5577 at 11 / 12 / 18 / 24 pt. `0.558` is the 12 pt value (residual < 0.01 pt); at 11 pt it would be ~0.06 pt off, just past the issue's 0.05 pt bar. Since this button is fixed at `size: 12` that is not a live concern, and the code comment says to re-measure if the anchor is ever reused at a very different size. I did not average the four rows, which would trade our exact size for sizes we do not draw.
- Left the constant private to `HeaderBar` rather than promoting it to a shared `Theme/` constant — there is one consumer, and `UserInterface` is views-only.
- No test target added: there is no `UserInterfaceTests` target and the issue explicitly does not ask for one.

## How to verify

**I could not run any of this.** The harness VM is Linux — no Swift toolchain (`swift: command not found`), no AppKit, no SwiftUI. Every number below is either quoted from the issue or derived from it; all three checks need a Mac or CI.

**1. Probe — pivot error before/after.** The issue's probe measures the glyph, which the patch does not change; what changed is the pivot, so the after-check is the residual against the anchor actually used. Replace the final `print` in the issue's probe with:

```swift
    let anchorY = 0.558                      // Self.refreshArcCenter.y
    print(String(format: "arc center y %.3f | box %.3f | old pivot dy %+.3f | new pivot dy %+.3f",
                 c.cy, Double(size.height),
                 c.cy - Double(size.height) / 2,
                 c.cy - anchorY * Double(size.height)))
```

Expected from the issue's own measurements: `old pivot dy +0.860`, `new pivot dy +0.000` — inside the ≤ 0.05 pt bar. Please confirm the run rather than the arithmetic.

**2. Probe — the rotation itself.** Stronger, and the check I would actually trust, since it exercises `rotationEffect(_:anchor:)` instead of re-deriving a constant: render the view *rotated* at several angles inside a fixed frame (so nothing clips) and fit the arc each time. The fitted center must not move.

```swift
MainActor.assumeIsolated {
    for deg in stride(from: 0.0, through: 315.0, by: 45.0) {
        let view = Image(systemName: "arrow.clockwise")
            .font(.system(size: 12, weight: .semibold))
            .rotationEffect(.degrees(deg), anchor: UnitPoint(x: 0.5, y: 0.558))
            .frame(width: 40, height: 40)
        guard let (_, all) = pixels(view, scale: 8) else { fatalError("render failed") }
        var p = all, c = fitCircle(p)
        for _ in 0..<6 {                     // same arrowhead trim as the issue's probe
            let res = p.map { abs((($0.0 - c.cx) * ($0.0 - c.cx)
                + ($0.1 - c.cy) * ($0.1 - c.cy)).squareRoot() - c.r) }
            let cut = res.sorted()[Int(Double(res.count) * 0.6)]
            let keep = zip(p, res).filter { $0.1 <= cut }.map { $0.0 }
            if keep.count < 20 { break }
            p = keep; c = fitCircle(p)
        }
        print(String(format: "%6.1f°  center (%.3f, %.3f)", deg, c.cx, c.cy))
    }
}
```

Before the patch (drop the `anchor:` argument) the printed center should trace a ~1.7 pt circle; after, it should be constant to well under 0.05 pt across all eight angles.

**3. App.** `cd native && xcodegen generate && open Tokcat.xcodeproj`, run, press Refresh with a real quota fetch in flight: the spin should read as in-place rotation, and the icon at rest must sit exactly where it does on `main` (compare against the gear's optical alignment).

**4. CI** must show green — `swift build`, `swift test`, and the xcodegen app build. `UserInterface` is not covered by either test target, so CI's job here is proving the file still compiles; `rotationEffect(_:anchor:)` and `UnitPoint` are both macOS 10.15+, well under the macOS 13 floor.

## Risks / follow-ups

- **Ink now sweeps ~0.86 pt lower.** Rotating about a pivot below the box center moves the swept disc down; the arrowhead can extend marginally further below the layout box than it used to. Nothing clips it (`.buttonStyle(.borderless)`, no `clipShape` on this path) and the swept radius about the true arc center is smaller than about the box center, so I expect no visible change — but it is the one thing worth a glance in the running app if the spin looks like it crowds the row.
- The constant is empirical. If SF Symbols reshapes `arrow.clockwise` in a future macOS, the anchor drifts silently — there is no test that would catch it. A `UserInterfaceTests` target running the probe above as an assertion would, if you ever want one; out of scope here.

Closes #86

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BCruLBzWpmdmH4ykLUMiV3

---
_Generated by [Claude Code](https://claude.ai/code/session_01BCruLBzWpmdmH4ykLUMiV3)_

### DIFF
--- a/native/LocalPackage/Sources/UserInterface/Views/HeaderBar.swift
+++ b/native/LocalPackage/Sources/UserInterface/Views/HeaderBar.swift
@@ -26,6 +26,21 @@ struct HeaderBar: View {
 
     @State private var spinStartedAt: Date?
 
+    /// Where `arrow.clockwise`'s arc center sits inside its layout box.
+    /// The arrowhead overshoots the ring at the top, so the ink box is
+    /// taller above the arc than below it and the default `.center` anchor
+    /// lands 0.86 pt above the circle the eye tracks (12 pt semibold) — the
+    /// glyph then orbits a 1.7 pt circle once per turn instead of spinning
+    /// in place. Measured by rendering this view with `ImageRenderer` and
+    /// least-squares fitting a circle to the ring pixels, arrowhead trimmed
+    /// as outliers: the center is horizontally centered (|dx| < 0.05 pt) and
+    /// vertically at 0.5535 / 0.5578 / 0.5591 / 0.5577 of the box height at
+    /// 11 / 12 / 18 / 24 pt — scale invariant, so one unit-space constant
+    /// serves. 0.558 is the value for the 12 pt we draw (residual < 0.01 pt);
+    /// re-measure if this ever gets reused at a very different size, and
+    /// expect only an SF Symbols redesign of the glyph to move it.
+    private static let refreshArcCenter = UnitPoint(x: 0.5, y: 0.558)
+
     /// The button drives both stores, so it has to keep spinning until the
     /// slower one lands — quota is the slower one (upstream calls plus
     /// subprocess spawns) and is usually why the user pressed Refresh.
@@ -79,9 +94,11 @@ struct HeaderBar: View {
                     TimelineView(.animation) { context in
                         Image(systemName: "arrow.clockwise")
                             .font(.system(size: 12, weight: .semibold))
-                            .rotationEffect(.degrees(
-                                context.date.timeIntervalSince(startedAt)
-                                    / 0.8 * 360))
+                            .rotationEffect(
+                                .degrees(
+                                    context.date.timeIntervalSince(startedAt)
+                                        / 0.8 * 360),
+                                anchor: Self.refreshArcCenter)
                     }
                 } else {
                     Image(systemName: "arrow.clockwise")


native/LocalPackage/Sources/UserInterface/Views/HeaderBar.swift
```diff
@@ -26,6 +26,21 @@ struct HeaderBar: View {
...
+    /// glyph then orbits a 1.7 pt circle once per turn instead of spinning
+    /// in place. Measured by rendering this view with `ImageRenderer` and
+    /// least-squares fitting a circle to the ring pixels, arrowhead trimmed
+    /// as outliers: the center is horizontally centered (|dx| < 0.05 pt) and
+    /// vertically at 0.5535 / 0.5578 / 0.5591 / 0.5577 of the box height at
+    /// 11 / 12 / 18 / 24 pt — scale invariant, so one unit-space constant
+    /// serves. 0.558 is the value for the 12 pt we draw (residual < 0.01 pt);
+    /// re-measure if this ever gets reused at a very different size, and
+    /// expect only an SF Symbols redesign of the glyph to move it.
+    private static let refreshArcCenter = UnitPoint(x: 0.5, y: 0.558)
```
handlecusion, line 42: **The whole fix is this number — check it hardest.**

`0.558` is the 12 pt row of the issue's table, not an average of the four rows. The measured ratios are 0.5535 / 0.5578 / 0.5591 / 0.5577 at 11 / 12 / 18 / 24 pt: an average (~0.557) would be marginally worse at the size we actually draw, so I optimized for our one call site. Consequence: at 11 pt this anchor would be ~0.06 pt off, just past the issue's 0.05 pt bar. That is why the doc comment says to re-measure before reusing it at a very different size, rather than presenting it as universal.

I could not run the probe (Linux VM, no Swift/AppKit) — `0.558 × 14.875 = 8.300` against the issue's measured `8.30` is arithmetic, not a measurement. Please run check 2 in the PR body, which fits the arc at eight rotation angles and would catch a wrong coordinate space, not just a wrong constant.

---

native/LocalPackage/Sources/UserInterface/Views/HeaderBar.swift
```diff
@@ -79,9 +94,11 @@ struct HeaderBar: View {
...
                         Image(systemName: "arrow.clockwise")
                             .font(.system(size: 12, weight: .semibold))
-                            .rotationEffect(.degrees(
-                                context.date.timeIntervalSince(startedAt)
-                                    / 0.8 * 360))
+                            .rotationEffect(
+                                .degrees(
+                                    context.date.timeIntervalSince(startedAt)
+                                        / 0.8 * 360),
+                                anchor: Self.refreshArcCenter)
```
handlecusion, line 101: Two things to confirm here, since I could not compile:

1. **The `else` branch is deliberately untouched.** At 0°/360° the rotation is the identity at *any* anchor, so the resting icon cannot move — that is the issue's "must not move" constraint satisfied by construction rather than by matching anchors across branches. Worth confirming visually anyway (check 3): the refresh icon at rest should be pixel-identical to `main`.

2. **`Self.refreshArcCenter` inside the `@ViewBuilder` closure.** `Self` resolves to `HeaderBar` here, and `UnitPoint` is `Sendable`, so a `private static let` is fine under the package's Swift 6 language mode. I am reasoning from the SDK rather than a compiler run — CI's `swift build` is the actual proof. If it does complain, the fallback is to inline the literal at this call site and keep the provenance comment above it.

---

handlecusion (commented):

Two-line diff, so the review surface is small. Where to look hardest, in order:

1. **The constant, `HeaderBar.swift:42`** — inline comment. It is the entire fix, it is empirical, and I chose the 12 pt row over an average of the four measured sizes.
2. **The `anchor:` argument, `HeaderBar.swift:97-101`** — inline comment. Covers why the non-spinning `else` branch stays as-is and the one Swift 6 concern I could not compile away.
3. **Nothing else rotates.** `grep -rn "rotationEffect" native/` returns only this call site, so there is no second glyph inheriting the same bug. Worth a second pair of eyes on that grep — if the tray animation or the 3D view rotates a symbol through some other API, I missed it.
4. **The verification I owe you is unrun.** The harness VM is Linux with no Swift toolchain and no AppKit, so neither probe nor the app ever executed here. The before/after numbers in the PR body are the issue's own measurements plus arithmetic. Please run check 2 (arc fitted at eight rotation angles) before merging — it is the one that would catch an anchor applied in the wrong coordinate space, which arithmetic cannot.
5. **The one behavioral risk**: the swept ink now sits ~0.86 pt lower in the row. I expect it invisible — the swept radius about the true arc center is *smaller* than about the box center, and nothing on this path clips — but it is the reason to glance at the row spacing while the spinner runs.

CI (`swift build`, `swift test`, xcodegen app build) is the compile proof; neither test target covers `UserInterface`, so CI cannot tell you the spin looks right.



---

handlecusion (commented):
🤖 **Claude Code** — 자동 리뷰

Nothing to flag.



---

native/LocalPackage/Sources/UserInterface/Views/HeaderBar.swift
handlecusion: 🤖 **Claude Code** — point 2 is now settled: **Swift build + test** passed on `a356af5` (macos-15, `swift build` + `swift test` + xcodegen app build). The `private static let refreshArcCenter = UnitPoint(...)` compiles under the package's Swift 6 language mode, so the inline-the-literal fallback is not needed. `Rust check` and `Frontend typecheck` are green too.

Point 1 still stands as a visual check for you — CI compiles the file but nothing in either test target renders `UserInterface`, so the resting-icon position and the spin itself are unverified by machine. Leaving this thread open for that reason.



---

PR #88, opened 2026-09-05
harness: merge under a dedicated PAT so the wrap-up event fires

handlecusion:
## Problem

Verified on #86 → #87: the gate's `Merged → close the linked issue and clear harness labels` step is **dead code**. `gh pr merge --auto` ran as `github-actions[bot]`, GitHub dispatches no workflow events for `GITHUB_TOKEN` actions, so `pull_request_target: closed` never fired — there is no such run in the workflow's history. The issue still closed (GitHub resolves `Closes #N` itself), but `claude` and `claude:pr-open` stayed on it.

## Change

- Merge under `HARNESS_MERGE_TOKEN`, a fine-grained PAT (resource owner `handlecusion`, repo `handlecusion/tokcat` only, Contents/Issues/Pull requests RW, expires 2027-09-06). A merge attributed to a real user fires the `closed` event.
- The PAT has **no Workflows permission** on purpose, so it cannot merge a PR that edits `.github/workflows/**` (this PR, for example). That path falls back to `GITHUB_TOKEN` and runs the wrap-up in-step, so approval is never a dead end. Same fallback when the secret is absent or expired (`::warning::` in the run log).
- Wrap-up extracted to `scripts/claude-harness/finish.sh` — idempotent, two call sites (in-step after an immediate merge, and the `closed` event for a deferred auto-merge). It now also drops the `claude` trigger label; leaving it on a closed issue made the documented re-run ("remove, then re-add") look like a no-op.
- Docs: `AGENTS.md` + `.claude/harness/README.md` (new secret, why it exists, and the new side effect that harness merges now trigger a `push: main` CI run).

## Verification

- `scripts/claude-harness/finish.sh 87` run locally against the already-merged PR #87: `#86 already closed, harness labels cleared` — idempotent, no spurious comment.
- YAML parses; every `run:` block passes `bash -n`.
- The end-to-end proof is the next harness PR: its merge must produce a `pull_request_target: closed` gate run whose wrap-up step is `success`, not `skipped`.

### DIFF
--- a/.github/workflows/claude-merge-gate.yml
+++ b/.github/workflows/claude-merge-gate.yml
@@ -13,6 +13,14 @@ name: Claude harness · merge gate
 # removed, the owner requests changes or dismisses their review, or new commits
 # are pushed — the owner re-approves what they actually reviewed.
 #
+# The merge itself runs under HARNESS_MERGE_TOKEN, a fine-grained PAT scoped to
+# this repo (contents + issues + pull-requests, read/write). It is not cosmetic:
+# GitHub dispatches no workflow events for anything GITHUB_TOKEN does, so a
+# github-actions[bot] merge never fires `pull_request_target: closed` and the
+# wrap-up below never runs — that is how #86 stayed labelled after merging. The
+# PAT deliberately has no Workflows permission, so a PR touching
+# .github/workflows/** falls back to GITHUB_TOKEN and wraps up in-step instead.
+#
 # A Claude cloud session acts under the owner's identity but *through the Claude
 # GitHub App*; a label applied that way is refused, so a session can never
 # approve its own work.
@@ -65,6 +73,10 @@ jobs:
       REVIEW_STATE: ${{ github.event.review.state || '' }}
       HARNESS_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
     steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
       - name: Decide
         id: decide
         run: |
@@ -110,6 +122,8 @@ jobs:
 
       - name: Enable auto-merge (merges when CI is green)
         if: steps.decide.outputs.mode == 'approve'
+        env:
+          MERGE_TOKEN: ${{ secrets.HARNESS_MERGE_TOKEN }}
         run: |
           state=$(gh pr view "$PR" --json state,isDraft,mergeStateStatus --jq '"\(.state) draft=\(.isDraft) merge=\(.mergeStateStatus)"')
           echo "PR #$PR: $state"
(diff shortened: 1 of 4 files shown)


.github/workflows/claude-merge-gate.yml
```diff
@@ -110,25 +122,60 @@ jobs:
...
+            local token=$1; shift
+            GH_TOKEN="$token" gh pr merge "$PR" "$@"
+          }
+          via=pat
+          if [ -z "${MERGE_TOKEN:-}" ]; then
+            echo "::warning::HARNESS_MERGE_TOKEN is not set — merging as github-actions[bot]; the closed event will not fire"
+            via=actions; MERGE_TOKEN="$GH_TOKEN"
           fi
+          # --auto is refused when nothing is pending (checks already green): merge directly.
+          if ! merge "$MERGE_TOKEN" --auto --merge && ! merge "$MERGE_TOKEN" --merge; then
```
handlecusion, line 151: `[bug]` The Workflows-permission fallback only covers the **immediate** merge path, but this PR is exactly the case that takes the deferred one.

`gh pr merge --auto` calls `enablePullRequestAutoMerge` — it *arms* auto-merge, it does not perform a merge. So the PAT's missing Workflows permission is never exercised by that call and it returns 0; the `&&` short-circuits, `--merge` is never tried, and the fallback block below is unreachable. The refusal instead lands later, when GitHub performs the merge on the PAT's behalf — outside this run, where nothing can fall back.

Net effect for a PR touching `.github/workflows/**` whose CI is still pending (#88 itself: `mergeable_state: blocked`): auto-merge is armed under a token that cannot complete it. The PR stays approved and unmerged, no `::warning::` is logged, `via` stays `pat` so the comment posts a plain "자동 머지 예약" with no caveat, and the deferred merge fails silently. That is the dead end the fallback was written to prevent.

Decide the token up front instead of by trial:

```sh
if gh api "repos/$GH_REPO/pulls/$PR/files" --paginate --jq '.[].filename' \
     | grep -q '^\.github/workflows/'; then
  echo "::warning::PR touches .github/workflows/** — merging as github-actions[bot]; the closed event will not fire"
  via=actions; MERGE_TOKEN="$GH_TOKEN"
fi
```

placed just after the empty-secret check, so those PRs take the documented `GITHUB_TOKEN` path (and its caveat comment) from the start. Worth confirming on the first live run that `--auto` really is accepted for the PAT here — if GitHub does validate the workflow scope at arm time, the current code is fine and only the ordering above is redundant.

---

.github/workflows/claude-merge-gate.yml
```diff
@@ -110,25 +122,60 @@ jobs:
...
+            [ "$via" = pat ] || how="$how (github-actions[bot]로 예약되어 머지 후 이슈 정리는 자동으로 돌지 않습니다.)"
           fi
           gh api -X POST "repos/$GH_REPO/issues/$PR/comments" \
             -f body="$(printf '✅ **Claude harness** — 오너 승인 확인 (%s). %s\n- 런: %s\n\n<!-- claude-harness kind=MERGE_GATE issue=%s -->' "$EVENT" "$how" "$HARNESS_RUN_URL" "$PR")" --silent
 
+          # Merged right here: no `closed` event is coming for this run's own
+          # merge when it went through GITHUB_TOKEN, and an idempotent extra
+          # wrap-up costs nothing when it does.
+          if [ "$merged" = yes ]; then
+            scripts/claude-harness/finish.sh "$PR"
```
handlecusion, line 176: `[bug]` Bootstrap: this resolves to **main's** copy of `finish.sh`, which does not exist yet.

The checkout added at L77 takes the default ref, which for both `pull_request_target` and `pull_request_review` is the base branch — main, not the PR head. (That is the right call for the trust boundary, and worth keeping.) But `scripts/claude-harness/finish.sh` is new in this PR, so `git show origin/main:scripts/claude-harness/finish.sh` is empty until #88 lands.

So if the owner approves after CI goes green, this run takes the immediate path: `--auto` refused (nothing pending) → `--merge` refused for the PAT (workflows) → `GITHUB_TOKEN` fallback merges → `merged=yes` → this line exits 127. The step fails *after* a successful merge, "Report failure on the PR" posts on a PR that in fact merged, and the wrap-up is skipped anyway.

One-time only — every later PR sees the script on main. Cheapest guard:

```sh
if [ "$merged" = yes ]; then
  if [ -x scripts/claude-harness/finish.sh ]; then
    scripts/claude-harness/finish.sh "$PR"
  else
    echo "::warning::finish.sh not on the base branch yet — skipping wrap-up"
  fi
fi
```

Alternatively land `finish.sh` on main in a separate PR first, then this one. Note the same line at L197 is safe: by the time a `closed` event fires, the merge has already put the script on main.

---

handlecusion (commented):
🤖 **Claude Code** — 자동 리뷰

Two problems on the fallback path, both hitting this PR itself rather than some future one.

The `.github/workflows/**` fallback is unreachable when CI is still pending: `--auto` only *arms* auto-merge, so the PAT's missing Workflows permission goes unexercised, the call succeeds, and the refusal lands later outside this run — #88 is `mergeable_state: blocked` right now, so this is the path it takes. Separately, the in-step `finish.sh` call runs from a main checkout that does not carry the script until this PR merges, so the immediate-merge path exits 127 after merging successfully.

Neither blocks review, but the end-to-end proof named in the description will not come from this PR's own merge unless one of them is addressed first.




---

PR #91, opened 2026-09-06
harness: make CI the session's compiler, explicitly

handlecusion:
Observed on #89 → #90: the first two rounds of that PR had never been compiled by anything, and the session said so in the PR body — the cloud VM is Linux, `native/` needs Swift + AppKit + SwiftUI. `ROUTINE_PROMPT.md` never mentioned that, and never told the session to wait for CI. The third round did poll CI on its own initiative and reported the run green, which is the behaviour this makes mandatory.

- New step 4 in DISPATCH (and required for every FOLLOW_UP push): poll the check runs for the head SHA, read the failing job's log, fix, push — up to 3 rounds — then hand off either green or with an explicit statement of what is red and why.
- Says outright that the session cannot build the project, so "it compiles" is never a claim it can make from reading.
- Says CI green is not visual verification, and asks for tests that measure the rendered artifact (`ImageRenderer` + ink measurement) over tests that restate a model the session just wrote — the tautology the auto-review caught on #90.
- `.claude/harness/README.md`: same, plus the note that a self-hosted macOS runner does not change this (the session VM is not swappable; only the runner's owner changes).

No workflow changes: `Swift build + test` already lands in ~2 min, so the existing CI is the loop.

Verification: prompt-only change. The next dispatched issue is the test — its PR should either be green at hand-off or name the red check. The clone's copy of `ROUTINE_PROMPT.md` takes precedence over the routine's stored snapshot, so this is live on merge; the snapshot at https://claude.ai/code/routines/trig_01QJ58u3U5nURAPzWJtXytGp still needs the same text pasted in for parity.


.claude/harness/ROUTINE_PROMPT.md
```diff
@@ -169,6 +173,25 @@ the PR body.
 - Comment on the issue with the PR link. Labels: add `claude:pr-open`, remove
   `claude:running` and `claude:needs-info`.
 
+### 4. Iterate against CI until it is green
+
+CI (`.github/workflows/ci.yml`, `macos-14`/`macos-15`) is the first machine that
+compiles your work. Treat a push as a build request, not a delivery:
+
+- After pushing, poll the check runs for your head SHA
+  (`list_workflow_runs` / `get_job_logs`, or `gh pr checks <PR> --watch`).
```
handlecusion, line 182: `[bug]` `gh pr checks <PR> --watch` contradicts ground rule 5 of this same file, which says: *"Never rely on `gh issue` / `gh pr` subcommands: the proxy serves only a pinned set of GraphQL operations"* — and, before that, that `gh` itself "is not in every environment" (`command -v gh`). So the one command offered here as the convenient path is the one the document already rules out twice. A session that reaches step 4 and reaches for it burns a round on a failing command at exactly the moment this step exists to save rounds.

`list_workflow_runs` / `get_job_logs` are fine, but the former is a *method* of `actions_list`, not a standalone tool, and neither is on rule 5's "verified in this environment" list — so a session may not find them quickly either.

Concrete fix — name the verified path and drop the `gh` one:

```
- After pushing, poll the check runs for your head SHA:
  `pull_request_read` (`method: get_check_runs`), then `get_job_logs`
  (`failed_only: true`) for the red job. `actions_list`
  (`method: list_workflow_runs`) also works.
```

`pull_request_read` is already on rule 5's verified list, and `get_check_runs` is keyed on the PR head commit, which is exactly what this bullet asks for.

---

.claude/harness/ROUTINE_PROMPT.md
```diff
@@ -169,6 +173,25 @@ the PR body.
...
+  `Swift build + test` takes ~2 minutes.
```
handlecusion, line 183: `[q]` `Swift build + test` is the only job named here, but `ci.yml` has **three** (`Frontend typecheck`, `Rust check`, `Swift build + test`) and per AGENTS.md the `protect-main` ruleset requires all three. A session that reads this step as "watch the Swift job" can hand off with that job green and `Rust check` red — a PR that still cannot auto-merge, which is the exact failure this PR is meant to eliminate.

Worth making the target explicit:

```
  `Swift build + test` takes ~2 minutes. Green means **all three** required
  checks (`Frontend typecheck`, `Rust check`, `Swift build + test`) — the
  ruleset gates on all of them, not just the Swift one.
```

Relevant in practice because AGENTS.md permits touching the Rust reader when the parity harness needs it.

---

handlecusion (commented):
🤖 **Claude Code** — 자동 리뷰

The new step 4 recommends `gh pr checks --watch`, which ground rule 5 of the same file explicitly forbids (`gh pr` subcommands are not served by the proxy, and `gh` may not be installed) — the polling step's headline command is the one the document already rules out. Worth fixing before merge, since a session hitting it loses a round in the loop this PR adds to save rounds.

Second, smaller: the step names only `Swift build + test`, but the ruleset gates on all three CI jobs, so "green" needs to mean all three.

Nothing here blocks merging on correctness grounds — both are wording fixes in a prompt-only change.



---

PR #95, opened 2026-09-12
Keep a Cmd-dragged menu-bar slot across rebuilds and relaunches

### LINKED ISSUE #94: Menubar item snaps back to its initial right-hand slot after the user Cmd-drags it left
Cmd-dragging the Tokcat item to the left of the menu bar doesn't stick: it later reappears at its initial slot near the right edge (between Wi-Fi and the clock area), next to the system items.

Observed on 0.3.3.

## Repro

1. Cmd-drag the Tokcat item left, e.g. right after the kiwi item (screenshot 1: `🥝 272 | 🐱 2.9M/m | ...`).
2. Keep using the machine (sleep/wake, undock, or just let it sit).
3. The item is back at the right end, after Wi-Fi (screenshot 2: `... 🔊 A 📶 | 🐱 2.8M/m`), i.e. roughly the 200 pt-from-right fallback slot.

## Suspected mechanism

Two independent paths in `native/LocalPackage/Sources/Model/AppShell/`, both worth checking before fixing:

**1. `StatusItemPlacement.decide` re-seats the fallback on the *second* rebuild after a drag.**

```swift
let userMoved = stored != nil && stored != seeded
if userMoved, !alreadyPreserved {
    return Decision(seatAt: nil, preservedUserPosition: true)
}
return Decision(seatAt: fallback, preservedUserPosition: false)
```

`preservedUserPosition` latches to `true` on the first hidden-item rebuild that
sees a user-moved position, so the *next* rebuild (`StatusItemController.recreateItem`,
≥120 s later per the backoff) takes the else branch and writes
`fallbackPosition = 200` into `NSStatusItem Preferred Position Item-0` — even though
the stored value is still the user's. The "the stored position is itself the problem"
escape hatch fires whenever the recovery heuristic triggers twice, not only when the
user's position is actually the cause of the hiding.

Note `checkVisibility` only requires two consecutive `isEffectivelyHidden` samples
(~30 s); a display sleep / undock / lid-close sequence can produce those without the
item being genuinely unreachable, so the rebuild path is not rare.

**2. The status item has no `autosaveName`.**

`StatusItemController` creates the item with
`NSStatusBar.system.statusItem(withLength:)` (lines 44 and 147) and never sets
`statusItem.autosaveName`, then calls `NSStatusBar.system.removeStatusItem` in
`deinit` and on every rebuild. Without an explicit autosave name AppKit owns the
generated `Item-0` slot and does not reliably keep the position across
remove/quit — so the drag can also be lost on plain relaunch, independent of
path 1. Setting a stable `autosaveName` (and keying the manual override off the
resulting `NSStatusItem Preferred Position <name>` default) is the idiomatic fix.

## Suggested direction

- Give the item a stable `autosaveName` and derive `positionKey` from it instead of hardcoding `Item-0`.
- Make the fallback re-seat fire only when preserving the user's position demonstrably failed to bring the item back (i.e. the rebuild happened *and* the item is still hidden on the next check), rather than latching on `alreadyPreserved`.
- Re-read the stored position after every rebuild so a drag performed between rebuilds re-arms the preserve branch.

## Tests

`native/LocalPackage/Tests/ModelTests/StatusItemPlacementTests.swift` pins the current
rule; it needs a case for "user moved the item again after a rebuild" and for
"two rebuilds must not reset a position the user still owns".

handlecusion:
🤖 **Claude Code**

## Summary

Fixes #94 — the item snapping back to the ~200 pt-from-right fallback slot after a Cmd-drag.

The mechanism is the one the owner established on-device on macOS 26 ([measurements](https://github.com/handlecusion/tokcat/pull/95#issuecomment-5647655793)), not the one this PR originally assumed:

**`NSStatusBar.removeStatusItem()` clears the item's saved position.** Hidden-item recovery removes and recreates the item, so every rebuild starts from an already-cleared placement key. `main`/0.3.3 reads `stored = nil`, concludes there is nothing of the user's to protect, and writes the fallback — the drag dies on the **first** rebuild. The `alreadyPreserved` latch this PR was originally built around was never involved.

That also means "preserve = write nothing" cannot work: it leaves the rebuilt item with no position at all, and the bar parks it off-screen until a later rebuild seats something. Measured on the first revision of this branch, that was a ~2 min disappearance — worse than the snap-back being fixed.

So the rule is now: **a rebuild always writes a position**, and the only question is whose. The user's own position is written back (which is also what repopulates the key for the next launch); the fallback wins only once that exact position has had a rebuild to itself and the item still hasn't come back.

## Changes

- **`StatusItemController.swift`**
  - `recreateItem()` reads `storedPosition` **before** `removeStatusItem()`; after that call there is nothing left to read.
  - `reseatPositionIfNeeded()` → `reseatPosition(stored:)`, which always writes.
  - `makeStatusItem()` sets a stable `autosaveName`; migration still runs in `init` before the first item exists.
  - `checkVisibility` clearing `preservedPosition` on the not-hidden path is unchanged — that is what stops an unrelated later rebuild from spending the user's slot.
- **`StatusItemPlacement.swift`** — `Decision.seatAt` is now a non-optional `Double`. Preserving a position means seating that exact value. New `seatsOurOwnValue`, which tells the controller whether the value may be recorded as Tokcat's.
- **`StatusItemPositionStore.swift`** — one source of truth. `storedPosition` reads the autosave key only; the legacy `Item-0` key is a migration input and is never written. `seat(at:)` (our fallback, stamps the seeded marker) is split from `restore(at:)` (the user's own position, deliberately does not). The migration is gated on its own marker key instead of "the autosave key is empty".
- **Tests** — the placement suite flips to the new `seatAt` expectations and keeps both #94 regressions; added "never seats anything but the stored position or the fallback" and a direct pin on `seatsOurOwnValue`. The store suite drops the two two-key-preference tests, and adds the legacy key not being a stored position, the migration being genuinely one-shot, and `restore` not claiming the value as ours.

## Assumptions

1. **The rebuilt item is placed from the key we just wrote.** This is the one link in the chain that no test here exercises: the measurements confirm macOS reads and writes `NSStatusItem Preferred Position TokcatStatusItem` with the autosave name set, and that it restores from it on launch, but "write the key between `removeStatusItem()` and the next `statusItem(withLength:)` and the new item lands there" is inferred from that, not observed.
2. **The stale `Item-0` value is left in place** rather than deleted after migration. It is never read again, and leaving it means a downgrade to 0.3.3 still finds a slot.
3. Exact `Double` comparison for positions, unchanged — AppKit writes back the value it read.

## How to verify

**CI** — `Swift build + test` on `macos-15` is the first machine that compiles this; nothing here was built locally (the session VM is Linux and has no Swift toolchain). The suites are pure logic over values and `UserDefaults`, so green is real coverage of the placement rule and the store — and of nothing above.

**Needs a Mac** (the owner's harness — CGEvent Cmd-drag, forced-rebuild hook, defaults poller — covers 1 and 2):

1. Real Cmd-drag left, then force one rebuild while the item is visible. The item must stay where it was dragged, and `defaults read com.handlecusion.tokcat "NSStatusItem Preferred Position TokcatStatusItem"` must still hold the dragged value. This is the assumption above, and the case the previous revision got wrong.
2. Parked seed (`3000` in the autosave key, `200` seeded): rebuild 1 re-seats 3000 and the item stays parked; rebuild 2 seats the fallback and it comes back.
3. Upgrade path: a machine with a dragged value under `…Item-0` and nothing under `…TokcatStatusItem` must not jump, and the new key must end up holding that number.

## Risks / follow-ups

- **Recovery from a genuinely bad drag takes two rebuilds, where `main` takes one** (measured: ~30 s → ~165 s). That is the direct cost of giving the user's position a chance before overruling it, which is what #94 asks for. The earlier claim that this was "no change in practice" was wrong.
- A position that is bad only on *some* display configs gets a fresh preserve-and-rebuild pair on each flip, i.e. one extra parked window per flip, rather than being remembered as bad. Remembering it the other way is what produced #94.
- Not addressed (out of scope for #94): `checkVisibility`'s two-sample heuristic can still fire without the item being genuinely unreachable. This change makes those false positives harmless instead of eliminating them — and the measurements found display sleep is *not* one of them, so the trigger in the original report is still unidentified.

Closes #94

### DIFF
--- a/native/LocalPackage/Sources/Model/AppShell/StatusItemController.swift
+++ b/native/LocalPackage/Sources/Model/AppShell/StatusItemController.swift
@@ -18,10 +18,6 @@ public final class StatusItemController: NSObject {
     private static let iconPointSize: CGFloat = 18
     private static let iconXOffset: CGFloat = 8
 
-    // Where macOS stores the item's spot, and the copy of the value we last
-    // wrote there ourselves (so a user drag is distinguishable from it).
-    private static let positionKey = "NSStatusItem Preferred Position Item-0"
-    private static let seededPositionKey = "TokcatSeededStatusItemPosition"
     private static let fallbackPosition = 200.0
 
     private var statusItem: NSStatusItem
@@ -33,7 +29,10 @@ public final class StatusItemController: NSObject {
     private var screenObserver: (any NSObjectProtocol)?
     private var hiddenChecks = 0
     private var lastRecoveryAt = Date.distantPast
-    private var preservedUserPosition = false
+    // Where macOS keeps the item's spot, and the user position the last rebuild
+    // left in place — cleared as soon as the item is seen on screen again.
+    private let positions = StatusItemPositionStore()
+    private var preservedPosition: Double?
 
     public var onLeftClick: (() -> Void)?
     public var menuProvider: (() -> NSMenu)?
@@ -41,7 +40,11 @@ public final class StatusItemController: NSObject {
     public var button: NSStatusBarButton? { statusItem.button }
 
     public override init() {
-        statusItem = NSStatusBar.system.statusItem(withLength: NSStatusItem.variableLength)
+        // Before the first item exists: AppKit reads the position preference
+        // when it places the item, so a pre-autosaveName drag has to be on the
+        // autosave key by then or the user's slot is lost for this launch.
+        StatusItemPositionStore().migrateLegacyPositionIfNeeded()
+        statusItem = StatusItemController.makeStatusItem()
         super.init()
         configureItem()
         startVisibilityWatch()
@@ -54,6 +57,17 @@ public final class StatusItemController: NSObject {
                 MainActor.assumeIsolated { self?.logDiagnostics() }
             }
         }
+    }
+
+    // A stable autosaveName is what keys the position preference to a name we
+    // control: AppKit's generated "Item-N" is positional, and it does not
+    // reliably survive removeStatusItem() or quit, so a Cmd-drag could be lost
+    // on a plain relaunch and on every hidden-item rebuild.
+    private static func makeStatusItem() -> NSStatusItem {
+        let item = NSStatusBar.system.statusItem(
+            withLength: NSStatusItem.variableLength)
+        item.autosaveName = StatusItemPositionStore.autosaveName
+        return item
     }
 
     private func configureItem() {
@@ -79,8 +93,8 @@ public final class StatusItemController: NSObject {
 
     // MARK: - Hidden-item recovery
     //
-    // macOS persists each status item's preferred position ("NSStatusItem
-    // Preferred Position Item-0"). After a display-layout change (undocking
+    // macOS persists each status item's preferred position (see
+    // StatusItemPositionStore). After a display-layout change (undocking
     // a notched MacBook, narrower bar) that position can land inside the
     // app-menu zone or the notch — the item's window is then parked far
     // off-screen and macOS NEVER re-places it, even when space frees up,
@@ -125,6 +139,11 @@ public final class StatusItemController: NSObject {
     private func checkVisibility() {
         guard isEffectivelyHidden else {
             hiddenChecks = 0
+            // The item is reachable on the position it currently has, so that
+            // position is not the cause of any later hiding. Forgetting it here
+            // is what keeps an unrelated second rebuild — display sleep, a
+            // dock/undock round trip — from reseating a slot the user chose.
+            preservedPosition = nil
             return
         }
         hiddenChecks += 1
@@ -144,8 +163,7 @@ public final class StatusItemController: NSObject {
         runner.removeFromSuperlayer()
         NSStatusBar.system.removeStatusItem(statusItem)
         reseatPositionIfNeeded()
-        statusItem = NSStatusBar.system.statusItem(
-            withLength: NSStatusItem.variableLength)
+        statusItem = Self.makeStatusItem()
         configureItem()
     }
 
@@ -156,16 +174,16 @@ public final class StatusItemController: NSObject {
     // system items, in the always-visible region. See StatusItemPlacement for
     // when the stored position is left alone instead.
     private func reseatPositionIfNeeded() {
-        let defaults = UserDefaults.standard
+        // Read fresh: a drag performed between two rebuilds must re-arm the
+        // preserve branch rather than look like the position we already tried.
         let decision = StatusItemPlacement.decide(
-            stored: defaults.object(forKey: Self.positionKey) as? Double,
-            seeded: defaults.object(forKey: Self.seededPositionKey) as? Double,
-            alreadyPreserved: preservedUserPosition,
+            stored: positions.storedPosition,
+            seeded: positions.seededPosition,
+            preserved: preservedPosition,
             fallback: Self.fallbackPosition)
-        preservedUserPosition = decision.preservedUserPosition
+        preservedPosition = decision.preserve
         guard let seatAt = decision.seatAt else { return }
-        defaults.set(seatAt, forKey: Self.positionKey)
-        defaults.set(seatAt, forKey: Self.seededPositionKey)
+        positions.seat(at: seatAt)
     }
 
     private var diagTimer: Timer?
--- a/native/LocalPackage/Sources/Model/AppShell/StatusItemController.swift
+++ b/native/LocalPackage/Sources/Model/AppShell/StatusItemController.swift
@@ -60,9 +60,11 @@ public final class StatusItemController: NSObject {
     }
 
     // A stable autosaveName is what keys the position preference to a name we
-    // control: AppKit's generated "Item-N" is positional, and it does not
-    // reliably survive removeStatusItem() or quit, so a Cmd-drag could be lost
-    // on a plain relaunch and on every hidden-item rebuild.
+    // control, instead of AppKit's generated and positional "Item-N". Measured
+    // on macOS 26: with the name set, the status bar reads and writes
+    // "NSStatusItem Preferred Position TokcatStatusItem" and leaves the
+    // generated key untouched — including when it restores the slot on launch,
+    // even though the name can only be assigned after the item is created.
     private static func makeStatusItem() -> NSStatusItem {
         let item = NSStatusBar.system.statusItem(
             withLength: NSStatusItem.variableLength)
@@ -99,10 +101,11 @@ public final class StatusItemController: NSObject {
     // app-menu zone or the notch — the item's window is then parked far
     // off-screen and macOS NEVER re-places it, even when space frees up,
     // while `isVisible` keeps reporting true. Reproduced deterministically
-    // by seeding a large preferred position. A freshly created item with no
-    // stored position always takes the first free slot on the right, so the
-    // recovery is: detect the parked state, drop the stored position, and
-    // rebuild the item (reusing the runner layer + stored title/state).
+    // by seeding a large preferred position. The recovery is: detect the parked
+    // state, rebuild the item (reusing the runner layer + stored title/state),
+    // and seat it — at the user's own position on the first try, at the
+    // fallback slot once that position has demonstrably failed to bring it
+    // back. See StatusItemPlacement.
 
     // Parked means the item's window sits outside every screen. Occlusion is
     // NOT a usable signal here: the menu bar legitimately disappears in
@@ -159,31 +162,39 @@ public final class StatusItemController: NSObject {
 
     private func recreateItem() {
         NSLog("Tokcat: status item parked off-screen; rebuilding it")
+        // Read before removing: removeStatusItem() clears the item's saved
+        // position, so after that call there is nothing left to preserve and
+        // the placement rule would only ever see nil.
+        let stored = positions.storedPosition
         appearanceObservation = nil
         runner.removeFromSuperlayer()
         NSStatusBar.system.removeStatusItem(statusItem)
-        reseatPositionIfNeeded()
+        reseatPosition(stored: stored)
         statusItem = Self.makeStatusItem()
         configureItem()
     }
 
-    // The fallback value is the distance from the RIGHT edge. Removing the
-    // preference doesn't help: a brand-new item is inserted at the LEFT end of
-    // the status area, which on a crowded bar is exactly the hidden zone it
-    // just died in (verified empirically). A small offset lands it beside the
-    // system items, in the always-visible region. See StatusItemPlacement for
-    // when the stored position is left alone instead.
-    private func reseatPositionIfNeeded() {
-        // Read fresh: a drag performed between two rebuilds must re-arm the
-        // preserve branch rather than look like the position we already tried.
+    // The fallback value is the distance from the RIGHT edge. Leaving the
+    // preference empty doesn't help: a brand-new item with no stored position
+    // is inserted at the LEFT end of the status area, which on a crowded bar is
+    // exactly the hidden zone it just died in (verified empirically). A small
+    // offset lands it beside the system items, in the always-visible region.
+    //
+    // A position is therefore always written — the removal above cleared the
+    // key — and the only question StatusItemPlacement answers is whether that
+    // position is the user's own or our fallback slot.
+    private func reseatPosition(stored: Double?) {
         let decision = StatusItemPlacement.decide(
-            stored: positions.storedPosition,
+            stored: stored,
             seeded: positions.seededPosition,
             preserved: preservedPosition,
             fallback: Self.fallbackPosition)
         preservedPosition = decision.preserve
-        guard let seatAt = decision.seatAt else { return }
-        positions.seat(at: seatAt)
+        if decision.seatsOurOwnValue {
+            positions.seat(at: decision.seatAt)
+        } else {
+            positions.restore(at: decision.seatAt)
+        }
     }
 
     private var diagTimer: Timer?
--- a/native/LocalPackage/Sources/Model/AppShell/StatusItemPlacement.swift
+++ b/native/LocalPackage/Sources/Model/AppShell/StatusItemPlacement.swift
@@ -5,26 +5,41 @@ import Foundation
 //
 // macOS stores the item's spot in "NSStatusItem Preferred Position <name>" —
 // the same preference it rewrites when the user Cmd-drags the item (see
-// StatusItemPositionStore). Hidden-item recovery used to overwrite that value
-// on every rebuild, which yanked the item back out of wherever the user had put
-// it. So: only reseat a value we wrote ourselves.
+// StatusItemPositionStore).
+//
+// The rebuild ALWAYS writes that preference, because the removal that starts
+// the rebuild clears it: measured on macOS 26, NSStatusBar.removeStatusItem()
+// leaves the item's saved position gone from the store. A rebuild that
+// "preserves" a position by writing nothing therefore hands the new item no
+// position at all, and the bar parks it off-screen until a later rebuild
+// seats something. So preserving the user's slot means seating that exact
+// value again, which is also what repopulates the key for the next launch.
 //
 // The one exception is a stored position that is itself the reason the item is
 // parked. That is only established when the *same* user position has already
 // had a rebuild to itself and the item never came back on screen since — which
 // is what `preserved` carries. A latching "we preserved once" flag was not
-// enough: the visibility watch also fires on display sleep and undock, so the
-// second rebuild it saw was usually an unrelated false positive, and it reset a
-// position the user still owned.
+// enough: the visibility watch also fires on unrelated relayouts, so the second
+// rebuild it saw was usually a false positive, and it reset a position the user
+// still owned.
 public enum StatusItemPlacement {
     public struct Decision: Equatable {
-        // The position to write, or nil to leave the stored one untouched.
-        public let seatAt: Double?
+        // The position to write. Always written — see above.
+        public let seatAt: Double
         // The user position this rebuild is giving a chance, carried into the
         // next decision. nil once the position has been overruled (we seated
         // the fallback); the caller also clears it as soon as the item is seen
         // on screen, which vindicates the position.
         public let preserve: Double?
+
+        // Whether `seatAt` is a value Tokcat picked rather than the user's own
+        // position being written back after the removal cleared it.
+        //
+        // Only our own value may be recorded in the seeded marker. Stamping a
+        // preserved user position as ours would make the *next* rebuild read it
+        // back as Tokcat's and drop the fallback on top of it — issue #94
+        // again, one rebuild later.
+        public var seatsOurOwnValue: Bool { preserve == nil }
     }
 
     public static func decide(
@@ -35,7 +50,7 @@ public enum StatusItemPlacement {
     ) -> Decision {
         guard let stored, stored != seeded else {
             // Nothing stored, or the value is one we seated ourselves:
-            // reseating it costs the user nothing.
+            // reseating the fallback costs the user nothing.
             return Decision(seatAt: fallback, preserve: nil)
         }
         if preserved == stored {
@@ -45,7 +60,8 @@ public enum StatusItemPlacement {
             return Decision(seatAt: fallback, preserve: nil)
         }
         // First rebuild for this position, or the user has dragged since the
-        // last one — give their choice a chance.
-        return Decision(seatAt: nil, preserve: stored)
+        // last one. Seat it again — that is what makes the rebuilt item
+        // reappear in the same slot instead of off-screen.
+        return Decision(seatAt: stored, preserve: stored)
     }
 }
--- a/native/LocalPackage/Sources/Model/AppShell/StatusItemPlacement.swift
+++ b/native/LocalPackage/Sources/Model/AppShell/StatusItemPlacement.swift
@@ -3,30 +3,49 @@ import Foundation
 // Where a rebuilt status item should sit. Split out of StatusItemController
 // (and kept free of AppKit) so the rule has direct test coverage.
 //
-// macOS stores the item's spot in "NSStatusItem Preferred Position Item-0" —
-// the same preference it rewrites when the user Cmd-drags the item. Hidden-item
-// recovery used to overwrite that value on every rebuild, which yanked the item
-// back out of wherever the user had put it. So: only reseat a value we wrote
-// ourselves. If preserving the user's choice didn't bring the item back, the
-// stored position is itself the problem and the fallback slot wins.
+// macOS stores the item's spot in "NSStatusItem Preferred Position <name>" —
+// the same preference it rewrites when the user Cmd-drags the item (see
+// StatusItemPositionStore). Hidden-item recovery used to overwrite that value
+// on every rebuild, which yanked the item back out of wherever the user had put
+// it. So: only reseat a value we wrote ourselves.
+//
+// The one exception is a stored position that is itself the reason the item is
+// parked. That is only established when the *same* user position has already
+// had a rebuild to itself and the item never came back on screen since — which
+// is what `preserved` carries. A latching "we preserved once" flag was not
+// enough: the visibility watch also fires on display sleep and undock, so the
+// second rebuild it saw was usually an unrelated false positive, and it reset a
+// position the user still owned.
 public enum StatusItemPlacement {
     public struct Decision: Equatable {
         // The position to write, or nil to leave the stored one untouched.
         public let seatAt: Double?
-        // Carried back into the next decision.
-        public let preservedUserPosition: Bool
+        // The user position this rebuild is giving a chance, carried into the
+        // next decision. nil once the position has been overruled (we seated
+        // the fallback); the caller also clears it as soon as the item is seen
+        // on screen, which vindicates the position.
+        public let preserve: Double?
     }
 
     public static func decide(
         stored: Double?,
         seeded: Double?,
-        alreadyPreserved: Bool,
+        preserved: Double?,
         fallback: Double
     ) -> Decision {
-        let userMoved = stored != nil && stored != seeded
-        if userMoved, !alreadyPreserved {
-            return Decision(seatAt: nil, preservedUserPosition: true)
+        guard let stored, stored != seeded else {
+            // Nothing stored, or the value is one we seated ourselves:
+            // reseating it costs the user nothing.
+            return Decision(seatAt: fallback, preserve: nil)
         }
-        return Decision(seatAt: fallback, preservedUserPosition: false)
+        if preserved == stored {
+            // This exact position already had a rebuild to itself and the item
+            // is parked again without ever having come back: the position is
+            // the problem, so the fallback slot wins.
+            return Decision(seatAt: fallback, preserve: nil)
+        }
+        // First rebuild for this position, or the user has dragged since the
+        // last one — give their choice a chance.
+        return Decision(seatAt: nil, preserve: stored)
     }
 }
--- a/native/LocalPackage/Sources/Model/AppShell/StatusItemPositionStore.swift
+++ b/native/LocalPackage/Sources/Model/AppShell/StatusItemPositionStore.swift
@@ -1,0 +1,76 @@
+import Foundation
+
+// The UserDefaults side of the menu-bar slot: which keys hold it, how a
+// pre-autosaveName install is migrated, and how a user's Cmd-drag is told
+// apart from a value Tokcat seated itself.
+//
+// macOS persists a status item's slot in "NSStatusItem Preferred Position
+// <autosaveName>" and rewrites that same preference when the user Cmd-drags
+// the item. Tokcat used to leave `autosaveName` unset and reach for AppKit's
+// generated "Item-0" name; that name is positional and is not reliably kept
+// across removeStatusItem()/quit, so a drag could be lost on a plain relaunch.
+// The item now carries a stable autosave name and this store owns the keys.
+//
+// Both keys are still read and written. The legacy value is migrated once so
+// upgrading users keep the slot they chose, and if a macOS version ignores our
+// autosave name and keeps maintaining the generated key, a fallback written
+// only to the new key would silently fail to move a parked item.
+public struct StatusItemPositionStore {
+    /// Stable autosave name for the status item. Also assigned to
+    /// `NSStatusItem.autosaveName`, which is what makes `positionKey` the key
+    /// macOS actually maintains.
+    public static let autosaveName = "TokcatStatusItem"
+
+    /// Where macOS stores the item's spot, derived from the autosave name.
+    public static let positionKey = "NSStatusItem Preferred Position \(autosaveName)"
+
+    /// The key AppKit generated for us before the autosave name existed.
+    public static let legacyPositionKey = "NSStatusItem Preferred Position Item-0"
+
+    /// The last position Tokcat seated itself; a stored value that differs
+    /// from it came from the user.
+    public static let seededPositionKey = "TokcatSeededStatusItemPosition"
+
+    private let defaults: UserDefaults
+
+    public init(defaults: UserDefaults = .standard) {
+        self.defaults = defaults
+    }
+
+    /// Copies a pre-autosaveName drag onto the autosave key.
+    ///
+    /// Must run *before* the first `NSStatusItem` is created: AppKit reads the
+    /// preference when it places the item, so a later write would not take
+    /// effect until the next launch.
+    public func migrateLegacyPositionIfNeeded() {
+        guard defaults.object(forKey: Self.positionKey) == nil,
+              let legacy = defaults.object(forKey: Self.legacyPositionKey) as? Double
+        else { return }
+        defaults.set(legacy, forKey: Self.positionKey)
+    }
+
+    /// The position macOS currently has on file. Read fresh on every rebuild so
+    /// a drag performed between two rebuilds is seen.
+    ///
+    /// A value that differs from `seededPosition` is by definition the user's,
+    /// so it wins over one we wrote — that is what picks the live key when only
+    /// one of the two is being maintained.
+    public var storedPosition: Double? {
+        let candidates = [Self.positionKey, Self.legacyPositionKey]
+            .compactMap { defaults.object(forKey: $0) as? Double }
+        let seeded = seededPosition
+        return candidates.first { $0 != seeded } ?? candidates.first
+    }
+
+    public var seededPosition: Double? {
+        defaults.object(forKey: Self.seededPositionKey) as? Double
+    }
+
+    /// Seats the item at `position` and records it as ours, so the next
+    /// decision knows the value is not a user drag.
+    public func seat(at position: Double) {
+        defaults.set(position, forKey: Self.positionKey)
+        defaults.set(position, forKey: Self.legacyPositionKey)
+        defaults.set(position, forKey: Self.seededPositionKey)
+    }
+}
--- a/native/LocalPackage/Sources/Model/AppShell/StatusItemPositionStore.swift
+++ b/native/LocalPackage/Sources/Model/AppShell/StatusItemPositionStore.swift
@@ -1,20 +1,21 @@
 import Foundation
 
-// The UserDefaults side of the menu-bar slot: which keys hold it, how a
+// The UserDefaults side of the menu-bar slot: which key holds it, how a
 // pre-autosaveName install is migrated, and how a user's Cmd-drag is told
 // apart from a value Tokcat seated itself.
 //
 // macOS persists a status item's slot in "NSStatusItem Preferred Position
 // <autosaveName>" and rewrites that same preference when the user Cmd-drags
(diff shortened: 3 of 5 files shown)


native/LocalPackage/Sources/Model/AppShell/StatusItemController.swift
```diff
@@ -56,6 +59,17 @@ public final class StatusItemController: NSObject {
...
     }
 
+    // A stable autosaveName is what keys the position preference to a name we
+    // control: AppKit's generated "Item-N" is positional, and it does not
+    // reliably survive removeStatusItem() or quit, so a Cmd-drag could be lost
+    // on a plain relaunch and on every hidden-item rebuild.
+    private static func makeStatusItem() -> NSStatusItem {
+        let item = NSStatusBar.system.statusItem(
+            withLength: NSStatusItem.variableLength)
+        item.autosaveName = StatusItemPositionStore.autosaveName
```
handlecusion, line 71: **Look here hardest — this is the one thing I could not verify by reading.** `autosaveName` is assigned *after* `statusItem(withLength:)`, which is the usual idiom, but AppKit reads the position preference when it places the item. If it only honours a name that was set before placement, then on this build the item would be placed using the generated `Item-0` slot and only pick up the autosave key on the *next* launch.

What I did to bound that risk: `migrateLegacyPositionIfNeeded()` (line 46) runs before this call so both keys hold the user's value on upgrade, and `seat(at:)` writes both keys so recovery can't silently no-op. But if verification step 2 in the PR body shows the item jumping once on upgrade, that confirms this timing and the fix is to swap which key is authoritative.

There is no AppKit API to create an item with an autosave name in one step, so this is as early as it can be set.

---

native/LocalPackage/Sources/Model/AppShell/StatusItemController.swift
```diff
@@ -125,6 +139,11 @@ public final class StatusItemController: NSObject {
     private func checkVisibility() {
         guard isEffectivelyHidden else {
             hiddenChecks = 0
+            // The item is reachable on the position it currently has, so that
+            // position is not the cause of any later hiding. Forgetting it here
+            // is what keeps an unrelated second rebuild — display sleep, a
+            // dock/undock round trip — from reseating a slot the user chose.
+            preservedPosition = nil
```
handlecusion, line 149: **This one line is the actual fix for the reported symptom.** Everything in `StatusItemPlacement` is downstream of it: without clearing here, a second rebuild still sees `preserved == stored` and spends the user's slot, exactly like the old boolean did.

Worth checking the reasoning: the claim is "the item is reachable on the position it currently has, therefore that position is not the cause of any *later* hiding". That holds for the notch/narrow-bar mechanism the recovery was built for (the position becomes invalid on a layout change, and a fresh layout change gets its own rebuild), but it does mean a position that is *intermittently* bad — reachable on one display config, parked on another — will get a fresh preserve+rebuild pair each time the config flips, i.e. one extra ~2 min parked window per flip instead of being remembered. I think that is the right trade: remembering it the other way is what produced #94.

---

native/LocalPackage/Sources/Model/AppShell/StatusItemPositionStore.swift
```diff
@@ -0,0 +1,76 @@
...
+        defaults.set(legacy, forKey: Self.positionKey)
+    }
+
+    /// The position macOS currently has on file. Read fresh on every rebuild so
+    /// a drag performed between two rebuilds is seen.
+    ///
+    /// A value that differs from `seededPosition` is by definition the user's,
+    /// so it wins over one we wrote — that is what picks the live key when only
+    /// one of the two is being maintained.
+    public var storedPosition: Double? {
```
handlecusion, line 71: **The most debatable design choice in the PR.** Reading two keys and picking "whichever differs from what we seated" is a heuristic, not a fact — it exists only because I can't confirm which key macOS actually maintains once `autosaveName` is set (see the comment on `makeStatusItem`).

Known edge: if the user drags the item *back onto* the fallback value, `storedPosition` equals `seededPosition` and the drag reads as ours, so a rebuild reseats it. That is a no-op write of the same number, so it's harmless — but it is the reason `treatsTheSeededValueAsOursEvenWhenPreserved` exists as a test rather than as a bug.

If step 2 of the manual verification confirms the autosave key is the live one, this whole property collapses to `defaults.object(forKey: Self.positionKey) as? Double` and `seat(at:)` stops writing the legacy key. I'd rather do that as a follow-up with the evidence than guess now.

---

native/LocalPackage/Sources/Model/AppShell/StatusItemPositionStore.swift
```diff
@@ -0,0 +1,76 @@
...
+    public init(defaults: UserDefaults = .standard) {
+        self.defaults = defaults
+    }
+
+    /// Copies a pre-autosaveName drag onto the autosave key.
+    ///
+    /// Must run *before* the first `NSStatusItem` is created: AppKit reads the
+    /// preference when it places the item, so a later write would not take
+    /// effect until the next launch.
+    public func migrateLegacyPositionIfNeeded() {
```
handlecusion, line 55: The upgrade path for every existing 0.3.3 user runs through here exactly once, so it's worth confirming the guard is the one you want: it copies only when the autosave key is **absent**, so a user who has already run a build with `autosaveName` and dragged since will never have that drag overwritten by the stale `Item-0` value. `migrationIsANoOpWithNothingStored` and `migrationNeverOverwritesTheAutosaveKey` pin both halves.

One thing this deliberately does *not* do is delete the legacy key after migrating — `seat(at:)` keeps writing it, so deleting it here would be inconsistent.

---

native/LocalPackage/Sources/Model/AppShell/StatusItemPlacement.swift
```diff
@@ -3,30 +3,49 @@ import Foundation
...
     ) -> Decision {
-        let userMoved = stored != nil && stored != seeded
-        if userMoved, !alreadyPreserved {
-            return Decision(seatAt: nil, preservedUserPosition: true)
+        guard let stored, stored != seeded else {
+            // Nothing stored, or the value is one we seated ourselves:
+            // reseating it costs the user nothing.
+            return Decision(seatAt: fallback, preserve: nil)
+        }
+        if preserved == stored {
```
handlecusion, line 56: The rule change in one line: the old code asked "have we ever preserved?", this asks "did *this exact position* already have a rebuild to itself, without the item coming back?".

Two consequences to sanity-check against your read of the bug:
- A drag between rebuilds makes `stored != preserved`, so preserve re-arms (`reArmsPreserveWhenTheUserMovesTheItemAgain`). That is suggestion 3 from the issue, and it falls out of re-reading `storedPosition` in `reseatPositionIfNeeded` rather than caching it.
- Genuine recovery is unchanged in timing: rebuild 1 preserves, the item stays parked, `preservedPosition` is never cleared, and the rebuild after the 120 s backoff seats the fallback. `fallsBackWhenPreservingDidNotRecoverTheItem` still covers that.

Exact `Double` equality is inherited from the code being replaced — AppKit writes the value it reads back, so this compares like-for-like rather than doing arithmetic on it.

---

handlecusion (commented):

Five places worth a hard look, inline. Ranked by how likely they are to be wrong:

1. **`makeStatusItem` / `autosaveName` timing** — the only claim in this PR I could not check by reading. Everything else degrades gracefully if I'm wrong here; this one decides whether path 2 of #94 is actually fixed.
2. **`storedPosition` reading two keys** — a heuristic that exists only because of (1). Collapses to one line once the timing is confirmed on a Mac.
3. **`preservedPosition = nil` in `checkVisibility`** — the actual fix for the symptom you reported. Short, but it's the whole behavioural change; the `StatusItemPlacement` rewrite only makes it expressible.
4. **`migrateLegacyPositionIfNeeded` guard** — runs once per existing install, so a wrong guard is a one-shot data loss for everyone on 0.3.3.
5. **`if preserved == stored`** — the rule change itself, and where to check that genuine recovery still gives up at the same point it used to.

Two honest limits: nothing here was compiled — the session VM is Linux, so `Swift build + test` on `macos-15` is the first machine that sees it — and the tests only cover the decision rule and the defaults plumbing. Whether AppKit honours the autosave name, and whether the item visually stays put across sleep/undock, needs a human at a Mac; steps 1–4 in the PR body are written to be run in that order, and step 2 (upgrade path) is the one that would invalidate the approach rather than just find a bug.



---

handlecusion:
🤖 **Claude Code** — CI green on the first push, no fix rounds needed.

- `Swift build + test` (macos-15) ✅ — 281 tests passed. Both new suites ran, not silently skipped: `StatusItemPlacementTests` and `StatusItemPositionStoreTests`, including `twoRebuildsDoNotResetAPositionTheUserStillOwns`, `reArmsPreserveWhenTheUserMovesTheItemAgain`, `prefersTheAutosaveKeyWhenThatIsWhereTheDragLanded`.
- `xcodegen + app build` ✅ — `BUILD SUCCEEDED`, lowercase `Contents/MacOS/tokcat` assertion passed.
- `Frontend typecheck` ✅, `Rust check` ✅.

**Green means it compiles and the logic tests pass — it does not verify this fix.** Nothing here ran on a menu bar. The two things CI cannot tell you, in the order I'd check them:

1. **Upgrade path** (`verify` step 2 in the PR body) — this is the one that could invalidate the approach rather than just find a bug. On a machine that already has a dragged value under `NSStatexItem Preferred Position Item-0`, does the item stay put, and does the new `…TokcatStatusItem` key end up holding the same number? A jump here means AppKit does not honour an `autosaveName` assigned after creation, and the fix is to flip which key is authoritative.
2. **The actual repro** (step 3) — drag left, then sleep/wake or undock a few times over ≥ 5 min.

Steps 1 and 4 (fresh install, and confirming recovery still self-heals from a seeded position of 3000) are cheap sanity checks after those two.



---

handlecusion:
🤖 **Claude Code** — typo above: the legacy key is `NSStatusItem Preferred Position Item-0` (I wrote `NSStatexItem`). The PR body has it right.



---

handlecusion (commented):
🤖 **Claude Code** — 자동 리뷰

`StatusItemPositionStore.swift:58-62` is the one to look hardest at. Once `migrateLegacyPositionIfNeeded()` (:45) copies a pre-upgrade drag onto `positionKey`, that copy differs from `seededPosition` and is read *first* — so on exactly the branch this two-key store exists for (macOS ignores `autosaveName` and keeps maintaining `Item-0`), every later drag lands in the legacy key and is shadowed by the frozen migrated value: `storedPosition` keeps returning the pre-upgrade position, the re-arm that `reArmsPreserveWhenTheUserMovesTheItemAgain` pins never fires, and the second rebuild writes `200` over a drag that never had its own rebuild — #94 again, for upgraders, until the first `seat(at:)` realigns both keys.

Read order alone can't fix it (the mirrored case — autosave key live, stale legacy drag left behind by an install that never seated — needs `positionKey` first). Recording what the migration copied under its own marker key and discounting a candidate equal to it, the same way `seeded` is discounted, covers both. A test for "migrated value present, drag arrives on the other key" would pin it.

[bug] above; nothing else blocks merging. Whether it can bite at all rides on the same unknown as the rest of the PR — if manual step 2 shows the autosave key is the live one, that branch reads correctly today.



---

handlecusion:
@claude — rework needed before this can merge. I verified this branch **on-device** on a real Mac, and the fix as written does not survive contact with the platform: it converts the reported symptom into a menu-bar item that disappears for a while. Everything below is measured, not inferred. The patch draft at the end is the shape I'd expect the fix to take.

## Method

macOS 26 (Darwin 25.5), MacBook Air M2, built-in Retina at 1680×1050 pt, Xcode 26.1.

- Built this branch (`0a4fad3`) and current `main` locally with the CI command.
- Drove a **real Cmd-drag** on the live `NSStatusItem` with CGEvent (key down → mouse down → drag → up), which is the same path a user gesture takes.
- Added a temporary `TOKCAT_FORCE_REBUILD=<seconds>` hook that calls `recreateItem()` directly, to reproduce the case the fix targets — a rebuild while the item is actually fine (the "false positive").
- Instrumented `recreateItem` / `reseatPositionIfNeeded` to print the placement keys both in-process and through a `/usr/bin/defaults` subprocess, so a UserDefaults cache artifact can't be mistaken for a change in the store.

## Findings

**1. Assumption 1 holds.** With `autosaveName` set, macOS reads and writes `NSStatusItem Preferred Position TokcatStatusItem`. Pinning that key at `700` while moving `…Item-0` through `300 / 1200 / 200` left the item window at literally the same `x=1019`; a real drag wrote the autosave key only (`…TokcatStatusItem=527`, later `625`) and left `…Item-0` untouched. The legacy key is neither read nor written — it is a migration source, never a second source of truth.

**2. Path 2 does not reproduce.** After a real drag, relaunch puts the item back at exactly the dragged offset — this branch (`rightOffset=527`) and 0.3.3 (`479`). The generated `Item-0` name is stable here: the first live item is always `Item-0`, and removing it hands `Item-0` back. I found no evidence for "the drag could be lost on a plain relaunch".

**3. Display sleep is not the trigger.** Over 4 minutes of `pmset displaysleepnow`, `isEffectivelyHidden` never fired: no rebuilds, `window.screen` stayed valid, only the occlusion bit flipped (8194 → 8192 → 8194).

**4. The mechanism this branch missed: `NSStatusBar.removeStatusItem(item)` clears the item's saved position.** Instrumented, three runs, same result:

```
before-remove: inProcess=Optional(479) external=479
after-remove:  inProcess=nil            external=            ← gone from the store
```

Same for a dragged value (`611`) and a parked one (`3000`). A minimal bundled probe running the same call did *not* reproduce it, so I could not isolate the exact AppKit condition — but in the app it is deterministic, and it is what drives findings 5–7. This is why `stored` is `nil` at the first rebuild; the `alreadyPreserved` latch is not involved.

**5. main/0.3.3 destroys the drag on the FIRST rebuild.** Real drag → `…Item-0=611` → one forced rebuild while the item is visible:

```
before-remove: Optional(611) → after-remove: nil → reseat: stored=nil seeded=200 → seatAt=200
window x=1035 → x=1368   (the ~200 fallback slot)
```

One rebuild. The "latch went true, then the *second* rebuild 120 s later wrote 200" story in the PR body is not what happens on this machine.

**6. This branch is worse in exactly that situation.** Real drag → `autosave=625` (and `legacy=479`, already stale by the drag) → one forced rebuild:

```
before-remove: autosave=Optional(625) legacy=Optional(479) seeded=Optional(200)
after-remove:  autosave=nil            legacy=Optional(479)
reseat: stored=Optional(479.0) seeded=Optional(200.0) preserved=nil
decision: seatAt=nil preserve=Optional(479.0)
```

The live key was cleared by the removal, the app reads the **stale mirror**, decides to "preserve", and writes nothing — so the recreated item has no position at all and the bar parks it off-screen:

```
02:42:18 x=991     (dragged)
02:42:21 x=-4327 screen=nil   ← item gone from the menu bar
02:42:45 rebuild #2: preserved == stored → seatAt=200
02:42:48 x=1368    (back, at the fallback slot)
```

On main the drag is lost but the item stays visible; here it disappears and only returns via a second rebuild. (My forced hook skips `checkVisibility`, so the second rebuild came ~25 s later; on the natural path the 120 s backoff applies, i.e. ~2 minutes off-screen.)

So the auto-review `[bug]` is not a narrow edge case: the stale mirror is the direct cause of the worst behaviour in this diff.

**7. Recovery timing is not "unchanged in practice".** Parked seed, both builds: main recovers in 1 rebuild (~30 s), this branch in 2 rebuilds (~165 s), because the mirror masks the cleared key. The `Risks` section says "no change in practice"; measurement says 30 s → 165 s.

## What the code must do

The whole design rests on "preserve = write nothing". That cannot work, because the key the item is placed from is cleared by the removal itself. Three coupled changes:

1. **Read the position before `removeStatusItem`** and pass it into the decision — after removal it is gone.
2. **Seat the preserved value** instead of writing nothing, so the recreated item has a position to be placed at (and the key is repopulated for the next launch).
3. **Drop the legacy read** (keep the one-shot migration for upgraders). A value that can go stale must never be what "the user's position" is derived from, and with 1+2 the mirror is unnecessary.

## Patch draft

`native/LocalPackage/Sources/Model/AppShell/StatusItemPlacement.swift` — preserve now means "re-seat this exact value", and `seatAt` is always decided:

```swift
public enum StatusItemPlacement {
    public struct Decision: Equatable {
        // Always written. Preserving a position means seating it again: the
        // system clears the item's saved position when the item is removed, so a
        // rebuild that writes nothing leaves the new item with no position and
        // the bar parks it off-screen.
        public let seatAt: Double
        // The user position this rebuild is giving a chance, carried into the
        // next decision. nil once the position has been overruled (we seated the
        // fallback); the caller also clears it as soon as the item is seen on
        // screen, which vindicates the position.
        public let preserve: Double?
    }

    public static func decide(
        stored: Double?,
        seeded: Double?,
        preserved: Double?,
        fallback: Double
    ) -> Decision {
        guard let stored, stored != seeded else {
            // Nothing stored, or a value we seated ourselves: reseating the
            // fallback costs the user nothing.
            return Decision(seatAt: fallback, preserve: nil)
        }
        if preserved == stored {
            // This exact position already had a rebuild to itself and the item
            // is parked again without having come back: the position is the
            // problem, so the fallback slot wins.
            return Decision(seatAt: fallback, preserve: nil)
        }
        // First rebuild for this position, or the user dragged since the last
        // one. Seat it again — that is what makes the rebuilt item reappear in
        // the same slot.
        return Decision(seatAt: stored, preserve: stored)
    }
}
```

`native/LocalPackage/Sources/Model/AppShell/StatusItemPositionStore.swift` — one source of truth; the legacy key is a migration input only:

```swift
    /// The key AppKit generated for us before the autosave name existed. Read
    /// once, by `migrateLegacyPositionIfNeeded()`, and never written: measured on
    /// macOS 26, once `autosaveName` is set the status bar neither reads nor
    /// maintains this key.
    public static let legacyPositionKey = "NSStatusItem Preferred Position Item-0"

    /// The position macOS currently has on file, read fresh on every rebuild.
    ///
    /// One key only. A two-key "whichever differs from what we seated" heuristic
    /// cannot tell a live value from a stale copy, and picking the stale one means
    /// the app seats a position the user never chose (which is how the item ended
    /// up parked off-screen in the measurements above).
    public var storedPosition: Double? {
        defaults.object(forKey: Self.positionKey) as? Double
    }

    public func seat(at position: Double) {
        defaults.set(position, forKey: Self.positionKey)
        defaults.set(position, forKey: Self.seededPositionKey)
    }
```

`native/LocalPackage/Sources/Model/AppShell/StatusItemController.swift` — capture before the removal, and let the decision always seat:

```swift
    private func recreateItem() {
        NSLog("Tokcat: status item parked off-screen; rebuilding it")
        // Read before removing: NSStatusBar.removeStatusItem() clears the item's
        // saved position, so after this call there is nothing left to preserve.
        let stored = positions.storedPosition
        appearanceObservation = nil
        runner.removeFromSuperlayer()
        NSStatusBar.system.removeStatusItem(statusItem)
        reseatPositionIfNeeded(stored: stored)
        statusItem = Self.makeStatusItem()
        configureItem()
    }

    private func reseatPositionIfNeeded(stored: Double?) {
        let decision = StatusItemPlacement.decide(
            stored: stored,
            seeded: positions.seededPosition,
            preserved: preservedPosition,
            fallback: Self.fallbackPosition)
        preservedPosition = decision.preserve
        positions.seat(at: decision.seatAt)
    }
```

The `checkVisibility` clear of `preservedPosition` on the not-hidden path stays as it is.

Tests that need to flip with it:

- `keepsAPositionTheUserDraggedTo` — `seatAt == 500`, `preserve == 500` (was `seatAt == nil`).
- `fallsBackWhenPreservingDidNotRecoverTheItem` — first rebuild `seatAt == 3000`, second `seatAt == fallback`.
- `twoRebuildsDoNotResetAPositionTheUserStillOwns` — both rebuilds `seatAt == 500`; that is the #94 regression test now, since the write-back is what keeps the slot.
- `reArmsPreserveWhenTheUserMovesTheItemAgain` — same shape, `seatAt` non-nil throughout.
- Store tests: keep key derivation + migration; drop `prefersAStoredValueThatIsNotTheOneWeSeated` and `prefersTheAutosaveKeyWhenThatIsWhereTheDragLanded`; add "a legacy value alone is not a stored position" (`storedPosition == nil` until the migration runs).
- Worth adding: a test that `decide` never returns a `seatAt` that differs from either the stored drag or the fallback.

## Verification I expect back

The symptom this PR is about is reproducible without a human, so please run it rather than reasoning about it:

1. Real Cmd-drag the item left, force one rebuild while it is visible → the item must stay where it was dragged, and `defaults read com.handlecusion.tokcat "NSStatusItem Preferred Position TokcatStatusItem"` must hold the dragged value afterwards.
2. Parked seed (`3000` in the autosave key, `200` seeded) → the item must still self-heal to the fallback after the second rebuild, i.e. recovery must not regress.
3. `swift test` + CI green.

I still have the harness for all three (CGEvent Cmd-drag, frame-aware drag helper, forced-rebuild hook, 4 Hz defaults poller) and can hand it over if a dev box is not available. I restored the test machine (defaults re-imported from the pre-test backup, the experimental autosave key removed, 0.3.3 relaunched).

native/LocalPackage/Sources/Model/AppShell/StatusItemController.swift
handlecusion: 🤖 **Claude Code** — settled by your measurements, and in the direction we needed. Finding 1: with the name set, macOS reads and writes `…TokcatStatusItem` and leaves `Item-0` untouched. Finding 2: relaunch restores from it. So a name assigned *after* `statusItem(withLength:)` is honoured for restoring, and the contingency this comment described — flip which key is authoritative — is off the table. Comment updated to say that as measured fact rather than an open risk.

What it does not settle is the narrower version of the same question, which is now assumption 1 in the PR body: that writing the key *between* `removeStatusItem()` and the next `statusItem(withLength:)` puts the new item there. That is verification step 1, and it is the one link in the chain no test here exercises.

The both-keys hedge this comment offered as the risk-bound is gone — see the `storedPosition` thread.



---

native/LocalPackage/Sources/Model/AppShell/StatusItemController.swift
handlecusion: 🤖 **Claude Code** — the line is unchanged and the trade-off reasoning in this comment still holds.

Two corrections your findings force, though. I justified this partly by display sleep producing false positives, and finding 3 shows display sleep never trips `isEffectivelyHidden` at all — the occlusion bit flips, `window.screen` stays valid, no rebuild. The clear is still right on its own terms (a position the item is currently reachable on is not the cause of a later hiding), but the actual trigger behind #94's repro is now unidentified rather than explained. I've moved that to an open follow-up instead of leaving it implied as understood.

And this is no longer "the actual fix for the reported symptom". With the removal clearing the key, writing the position back on every rebuild is what keeps the slot; this line only stops the *second* rebuild from spending it. On `main` the drag was already gone by then — finding 5.



---

native/LocalPackage/Sources/Model/AppShell/StatusItemPositionStore.swift
handlecusion: 🤖 **Claude Code** — deleted, exactly as this comment said it would be if the autosave key turned out to be the live one. It now reads `defaults.object(forKey: Self.positionKey) as? Double`, and `seat(at:)` no longer writes the legacy key.

Worth recording that it was worse than the redundancy this comment anticipated. Finding 6: after the removal cleared the live key, "whichever differs from what we seated" had only the stale `Item-0` copy left to pick, so the app took a value the user had already dragged away from and called it the user's position — then "preserved" it by writing nothing, and the item went off-screen. The heuristic was not a harmless hedge waiting for evidence; it was the direct cause of the worst behaviour in the diff, which is what the auto-review's `[bug]` was pointing at.

The edge this comment documented survives unchanged: dragging back onto the fallback value still reads as ours, still a no-op write, still covered by `treatsTheSeededValueAsOursEvenWhenPreserved`.



---

native/LocalPackage/Sources/Model/AppShell/StatusItemPositionStore.swift
handlecusion: 🤖 **Claude Code** — you asked me to confirm the guard is the one I wanted, and finding 4 says it isn't, on its own.

"The autosave key is absent" was standing in for "we have never migrated". A removal empties that key, so a launch starting from a cleared one would copy the stale `Item-0` value back over a newer drag — the migration would not be one-shot at all. It is now gated on its own marker (`TokcatDidMigrateLegacyStatusItemPosition`), with the never-overwrite condition kept as a second guard so both halves you named still hold. `migrationRunsOnlyOnceEvenIfTheAutosaveKeyIsClearedLater` pins the new one; `migrationNeverOverwritesTheAutosaveKey` and `migrationIsANoOpWithNothingStored` are unchanged.

The "deliberately does not delete the legacy key" note stands, but not for the reason given here — `seat(at:)` no longer writes it, so consistency is no longer the argument. It is that the key is never read again after this runs, and leaving it means a downgrade to 0.3.3 still finds a slot.



---

native/LocalPackage/Sources/Model/AppShell/StatusItemPlacement.swift
handlecusion: 🤖 **Claude Code** — the rule itself is unchanged: "did *this exact position* already have a rebuild to itself, without the item coming back?". Both consequences listed here still hold, and `reArmsPreserveWhenTheUserMovesTheItemAgain` still covers the first.

What changed is the other branch. It returns `seatAt: stored` rather than `seatAt: nil`, because writing nothing leaves the rebuilt item with no position at all. `Decision.seatAt` is non-optional now.

One claim in this comment has to be withdrawn: "genuine recovery is unchanged in timing" was wrong. Measured, it is 1 rebuild on `main` and 2 here — ~30 s → ~165 s. The PR body now states that as the price of giving the user's position a chance rather than denying there is one.

There is also a trap in the re-seat that the rule alone doesn't cover, so it's handled in the store: seating a *preserved* position must not stamp the seeded marker. If it did, the next rebuild would see `stored == seeded`, take the first branch, and drop the fallback on a slot the user still owns — #94 again, one rebuild later, with every test here still green because they pass `seeded:` by hand. Hence `restore(at:)` next to `seat(at:)`, chosen by `seatsOurOwnValue`, with `claimsOnlyTheFallbackAsOurs` and `restoringAUserPositionDoesNotClaimItAsOurs` pinning the two halves.



---

handlecusion:
🤖 **Claude Code** — reworked to your patch draft in `d551149`. CI green. The verification you asked for, I could not run; details at the bottom, and it is the part that matters most here.

## What changed, per your three points

1. **Read before the removal** — `recreateItem()` captures `positions.storedPosition` before `NSStatusBar.system.removeStatusItem(statusItem)` and passes it into `reseatPosition(stored:)`.
2. **Seat the preserved value** — `Decision.seatAt` is a non-optional `Double`; preserving means writing that exact value back. `reseatPositionIfNeeded()` → `reseatPosition(stored:)`, since it no longer has an "if needed".
3. **Drop the legacy read** — `storedPosition` is `defaults.object(forKey: Self.positionKey) as? Double`. `Item-0` is a migration input, never written.

## One place I did not follow the draft

`seat(at:)` stamps the seeded marker, and the draft routes the preserved value through it. That re-breaks #94 one rebuild later:

```
drag 500, seeded 200
rebuild 1: stored=500 ≠ seeded=200 → seatAt=500, preserve=500 → seat(500) → seeded becomes 500
item comes back; checkVisibility clears preservedPosition
rebuild 2: stored=500 == seeded=500 → first branch → seatAt=200
```

The item lands on the fallback again, from a rebuild that had nothing wrong with it. It does not show up in the suite because the placement tests pass `seeded:` by hand and never see the marker the controller actually writes — including `twoRebuildsDoNotResetAPositionTheUserStillOwns`, which passes either way.

So the store has two writes: `seat(at:)` (our fallback — writes the key and claims it) and `restore(at:)` (the user's position — writes the key only). `Decision.seatsOurOwnValue` picks between them. Pinned by `claimsOnlyTheFallbackAsOurs` and `restoringAUserPositionDoesNotClaimItAsOurs`.

One other deviation: the migration is now gated on its own marker key. "The autosave key is absent" was standing in for "never migrated", and finding 4 breaks that — a launch starting from a key some removal had cleared would copy the stale `Item-0` value back over a newer drag. The never-overwrite guard is kept as well.

## Tests

Flipped as you listed. Dropped `prefersAStoredValueThatIsNotTheOneWeSeated` and `prefersTheAutosaveKeyWhenThatIsWhereTheDragLanded`; added `aLegacyValueAloneIsNotAStoredPosition`, `onlyEverSeatsTheStoredPositionOrTheFallback` (your suggestion), plus the two marker tests above, `migrationRunsOnlyOnceEvenIfTheAutosaveKeyIsClearedLater`, and `restoringRepopulatesAClearedKey`.

`Swift build + test` ✅ 285 tests (was 281), all four new ones plus the reshaped store suite ran by name, not skipped. `xcodegen + app build` ✅ `BUILD SUCCEEDED`, lowercase executable assertion passed. `Frontend typecheck` ✅, `Rust check` ✅.

## The verification you asked for

**I did not run items 1 and 2, and I am not going to claim otherwise.** This session's VM is Linux with no Swift toolchain at all — not merely no Xcode, so `swift test` does not run here either; CI was the only compiler in this loop, and it never touches a menu bar. Please take the harness offer, or point it at this branch yourself.

Item 1 is the one that matters, because the whole design now rests on a link nothing here tests: that writing the key **between** `removeStatusItem()` and the next `statusItem(withLength:)` is what the new item is placed from. Your findings 1 and 2 establish that macOS maintains and restores from the autosave key — they do not establish that a write in that window takes effect on the item created right after it. If it doesn't, the item is off-screen again and the fix needs a different insertion point, not a different rule. That is assumption 1 in the PR body.

Item 2 should hold by construction (rebuild 1 re-seats 3000 and the item stays parked, rebuild 2 sees `preserved == stored` and seats 200), but it is worth confirming that the re-seat of 3000 doesn't change the parked geometry in some way the logic can't see.

I also corrected the PR body: Path 2 is gone (finding 2 — it does not reproduce), and the "no change in practice" line on recovery timing is replaced with the measured ~30 s → ~165 s and why it is the price of #94. Separately, since finding 3 rules out display sleep, the trigger behind the original repro is now unidentified — I've left that as an open follow-up rather than implying this PR explains it.



---

handlecusion:
**Verification run on-device against `d551149` — all three items from "How to verify", plus the case the previous revision got wrong. Passing.**

Harness: same as [the rework brief](https://github.com/handlecusion/tokcat/pull/95#issuecomment-5647655793) — local build of this revision with a temporary `TOKCAT_FORCE_REBUILD` hook, real Cmd-drags via CGEvent, `TOKCAT_DIAG_STATUS` window frames, and the placement keys read both in-process and through `/usr/bin/defaults`.

**1. Drag survives a rebuild while the item is visible (the #94 case).** Seeded `479`/`seeded 200`, real Cmd-drag to `561`, forced rebuild at t=25 s:

```
before-remove: stored=Optional(561.0) seeded=Optional(200.0) preserved=nil
after-remove:  autosave=nil  seeded=Optional(200)
decision:      seatAt=561.0 preserve=Optional(561.0) writePath=restore
x=1021 w=98  →  right offset 561, before and after
```

The item never left the bar (no `screen=nil` frame), `…TokcatStatusItem` still holds `561` afterwards, and `TokcatSeededStatusItemPosition` is still `200` — the preserved value was not recorded as ours.

**2. Two consecutive rebuilds.** Same setup, forced rebuilds at t=25 s and t=50 s (`preserved` cleared by the visible sighting in between, which is the case that matters):

```
rebuild 1: stored=561 seeded=200 preserved=nil → seatAt=561 restore
rebuild 2: stored=561 seeded=200 preserved=nil → seatAt=561 restore
```

Both rebuilds re-seat `561`; neither falls through to the fallback. This is where the draft in my comment was wrong — routing the preserved value through `seat(at:)` would have stamped the seeded marker and made rebuild 2 read `561` back as Tokcat's own. `restore(at:)` with `seatsOurOwnValue` is doing what it says.

**3. Recovery still self-heals.** Parked seed (`3000`/`200`):

```
rebuild 1 (t=30s):  stored=3000 preserved=nil    → seatAt=3000 restore  → item stays parked
rebuild 2 (t=164s): preserved==stored            → seatAt=200  seat     → x=1355, visible
final keys: autosave=200, seeded=200
```

One backoff cycle, as the body now documents. vs 0.3.3's single rebuild this is the ~30 s → ~165 s trade, and it is worth paying: 0.3.3 gets there by destroying the drag on rebuild 1.

**4. Upgrade / migration paths (item 3)**, key-level results:

- legacy `Item-0=700`, no autosave key, no marker → migration copied **once** (`autosave=700`, marker set), and the item is placed from it.
- newer drag `900` with a stale `Item-0=700` and the marker set → `autosave` stays **900**; the stale value is not resurrected.
- autosave key cleared by a removal, marker set, stale `Item-0=700` → **no copy**; the key stays absent. (This is the case the marker key was added for — "the autosave key is empty" would have resurrected `700` over a newer drag.)

**5. Relaunch persistence.** A drag-`479` state survives restart: `rightOffset=481` on both launches.

**6. Locally reproduced CI.** `swift build && swift test` on `d551149`: **285 tests / 48 suites pass**. Matches the session's report. Required checks on `d55114993` are all `success`.

### Two things worth recording

- **Assumption 1 is now observed, not inferred.** The write between `removeStatusItem()` and the next `statusItem(withLength:)` is what the new item is placed from: in runs 1 and 2 the rebuilt item lands on exactly the offset we wrote, and the previous revision's off-screen outcome is explained by not writing at all. You can move that line out of `Assumptions`.
- **Absolute positions are bar-state dependent, and that is not a code problem.** A placement matrix on this menu bar right now: `200→226`, `400→416`, `500→529`, `600→639`, `700→parked`, `900→parked`. So a value that was placeable earlier can be invalid later, and restoring such a value parks the item — after which the app recovers to the fallback by design (run 3). Expect that shape in any log you read.

### What I could not do

The *natural* trigger for the original repro is still unidentified on this machine: 4 minutes of display sleep produces no rebuild at all (`isEffectiveHidden` never fires), and a left Cmd-drag is clamped by the system into a placeable slot, so it never parks the item. Runs 1, 2 and 4 therefore use the forced-rebuild hook — the same `recreateItem()` path `checkVisibility` calls, minus the heuristic. The PR body already says the trigger is an open follow-up; nothing here changes that.

No blockers found. Verification authorises merging from my side; the owner approval is still yours. `#94` should stay open until this ships (0.3.4 patch, per AGENTS.md).

Test machine restored: defaults re-imported from the pre-test backup (and the two experimental keys removed), 0.3.3 relaunched, worktree removed, checkout clean.

handlecusion:
**Addendum — the real-Cmd-drag coverage, item by item.** Following up on my verification comment: item 3 (upgrade path) is now verified with an actual drag rather than key-level seeding, and the relaunch path is closed for a placeable value. One new lead on the original trigger at the bottom.

**Upgrade path with a real drag (your item 3).** 0.3.3, fresh state, real Cmd-drag → `Item-0=307`, item visible at `rightOffset=307`. Quit, launch this revision:

```
upgraded launch:      rightOffset=307   (no jump)   autosave=307   marker=1
forced rebuild t=35:  stored=307 seeded=200 preserved=nil → seatAt=307 writePath=restore
                      rightOffset=307   autosave=307   seeded=200
```

The migration copied the dragged value once, the item did not move, and the drag then survived a rebuild on the new code. Previously I had only verified the two halves separately.

**Real drag → relaunch, placeable value.** Drag `400 → 480`, quit, relaunch: `rightOffset=480` and `autosave=480` both times. Drag to the leftmost slot the bar currently accepts (`626`), relaunch: `rightOffset=624`, stable for 200 s with **no rebuilds at all**.

**Real drag → rebuilds.** Already reported: one rebuild keeps `639`, two consecutive rebuilds keep `561`, in both cases `seeded` stays `200` and the item never leaves the bar.

### New lead on the original #94 trigger (follow-up, not a blocker)

A real drag can store a value the bar accepts at drop time and **refuses at item creation**. I hit it by accident: a drag off an item that had been created with no stored position produced `…TokcatStatusItem=964` with the item visibly placed, and every subsequent launch from `964` parks the item off-screen (`rightOffset≈5642`, reproduced twice, deterministic). The app then does what it is supposed to — two rebuilds, then the fallback — i.e. the user sees *"I dragged it left, and later it is back near the right edge"* with no false positive anywhere in the chain.

That is worth chasing separately from this PR: it suggests part of #94's report can be a slot that the bar will not restore, not a destroyed drag. It does not change anything here — the fix's job is to not spend a position the user owns, and it now also restores genuinely invalid positions to a visible slot after one backoff cycle.

Context for reading logs: the launch-placeable band on this machine is bar-state dependent. A matrix taken now: `200→226`, `400→416`, `500→529`, `600→639`, `700→parked`, `900→parked`, `964→parked`. Drags can reach further left than creation can restore, which is exactly the lead above.

No change to the conclusion: no blockers found on my side, owner approval is the remaining gate. Machine restored again (defaults re-imported from the pre-test backup, the two experimental keys removed, 0.3.3 relaunched).

PR #96, opened 2026-09-12
Measure the three cloud-session usage sources before wiring one into the graph

handlecusion:
Claude Code cloud sessions and Codex Cloud tasks are showing up more often and neither is visible in the graph today. This PR measures both: which source is accurate, what it costs, and — after chasing the Codex side down — **what each can actually yield**. No app behavior changes yet; the decision comes next.

## Codex Cloud: reachable, but percent — not tokens

Started from the two dead ends:

| probe | result |
| --- | --- |
| `codex cloud list --json` | exit 0, 782 ms, **4 tasks**, fields `attempt_total, environment_id, environment_label, id, is_review, status, summary, title, updated_at, url` — **no usage field** |
| `POST /wham/usage/thread_usage/query` (per-thread usage) | **403** with the CLI's OAuth token (v2 → 404); adding the app's `Origin`/`Referer`/beta headers does not change it |
| `GET /wham/tasks/{id}/turns/{turn}/logs` | `{"logs": []}` |
| `GET /wham/usage/credit-usage-events` | `{"data": []}` |
| `GET /wham/usage/plan_limit_history` | 404 |
| `GET /wham/usage/daily-workspace-user-token-usage-breakdown` | 400 — `No active workspace found for the current account` (business/workspace variant) |

The route list came from ChatGPT.app's `app.asar`, which also gave the call shape the app itself uses:

```js
safeGet('/wham/usage/daily-token-usage-breakdown',
  {parameters: {query: {start_date, end_date, group_by: 'day'}}})
```

That one works with local credentials, and it is implemented here as `CodexCloudUsageProvider`:

| measurement | value |
| --- | --- |
| wall time | **869 ms** (1 GET, existing credential + refresh path) |
| response | 90 days for a 90-day range; `units: "percent"`, `group_by: "day"` |
| shape | `data[].product_surface_usage_values` — per-day plan percent **split by where the work ran** |
| cloud surfaces | `web`, `work_web`, `mobile`, `work_mobile`, `slack`, `linear`, `jetbrains`, `github`, `github_code_review`, `agent_identity`, `sdk` |
| local surfaces | `cli`, `vscode`, `desktop_app`, `exec`, `unknown` (this Mac — Tokcat already parses these from disk) |
| 90-day totals on this account | cloud **0.5112 %**, local **135.367 %** |
| cloud non-zero days | **1** — 2026-09-05: `web` 0.024 %, `agent_identity` 0.487 % |

So Codex Cloud usage **is** retrievable per day, and the cloud/local split is exactly what a cloud row would need — but the unit is plan percent. Reading it as tokens is not defensible: calibrating "tokens per percent" against days where both the account's local percent and Tokcat's local token count exist gives

| day | local percent | local tokens (Tokcat) | tokens / percent |
| --- | --- | --- | --- |
| 2026-06-18 | 0.243 | 1,224,770 | 5.04 M |
| 2026-09-05 | 31.972 | 4,092,946 | 0.128 M |
| 2026-09-06 | 100.000 | 293,018,446 | 2.93 M |
| 2026-09-07 | 1.374 | 181,697,807 | 132.2 M |
| 2026-09-11 | 1.702 | 5,351,981 | 3.14 M |

a **1000× spread** across five days (and 2026-09-06 sits at exactly `100.000`, so the percent is bounded/capped rather than a linear share). Any "cloud tokens" figure derived from it would be invention.

## Claude cloud: tokens, but almost nothing on disk

The updated marker × read matrix (ground truth = `cwdField` + `fullScan`; 1,093 candidate transcripts, 472.9 MB):

| marker | read | files | false pos | precision | recall | MB read | probe wall | tokens attributed |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| `pathSlug` | none | 1 | 0 | 1.00 | 1.00 | **0.0** | **0.21 s** | 16,971 |
| `cwdField` | head 8 KB | 1 | 0 | 1.00 | 1.00 | 6.7 | 0.38 s | 16,971 |
| `cwdField` | head 64 KB | 1 | 0 | 1.00 | 1.00 | 50.4 | 1.2 s | 16,971 |
| `cwdField` | full scan | 1 | 0 | 1.00 | 1.00 | 473.4 | 6.9 s | 16,971 |
| `rawSubstring` | full scan | 4 | **3** | **0.25** | 1.00 | 473.4 | 11.2 s | 67.1 M (56.6 M wrong) |
| `bridgeSessionID` | full scan | 55 | **55** | **0.00** | 0.00 | 473.4 | 6.0 s | 406.2 M |

Two traps the tests now pin: raw substrings catch transcripts that merely *print* the worktree path (`git worktree list` output, an `ls` error), and `bridgeSessionId: cse_…` appears on 55 ordinary locally bridged sessions, not just cloud ones. Parsing the `cwd` **value** removes both.

Coverage is the real limit: exactly one cloud transcript is visible locally (2026-08-30, **16,971 tokens = 0.022 %** of that day's 78.6 M claude tokens), against **10** cloud sessions this repo's harness has fired. A transcript only exists when a Remote Control bridge was registered at fire time.

## Yield summary

| source | what it yields | accuracy | cost |
| --- | --- | --- | --- |
| Claude local transcripts (`pathSlug`) | real tokens, daily — 16,971 tokens, 1 session | exact, precision/recall 1.00 locally; ~10 % of cloud sessions reach disk | 0 bytes read, +0.21 s |
| Codex `daily-token-usage-breakdown` | plan **percent** per day, cloud/local split — 1 cloud day in 90 | exact as served; not tokens (1000× calibration spread) | 869 ms, 1 GET + credential path |
| Codex `cloud list --json` | task inventory only (4 tasks) | exact, no usage | 782 ms, subprocess + network |
| percent → tokens estimate | nothing defensible | **rejected** | — |

Recommendation: wire the Claude split into the graph (`pathSlug`, exact, free) so bridged cloud sessions stop hiding inside `claude`, and surface the Codex cloud percent where plan percent already lives (the limits card) — not in the token graph, where it would have to be invented. If the per-thread token endpoint ever opens up (`thread_usage/query_v2` returning instead of 403/404), that becomes the Codex token source; the probe already prints what it returns.

## Verification

- `swift build && swift test` — **281 tests, 49 suites**, pass (`CloudSessionMarkerTests`, `CodexCloudUsageTests`).
- `scripts/parity-check.sh --snapshot` — **PASS** (collector stays bit-compatible with the Rust side).
- Reproduce: `cd native/LocalPackage && swift build && ./.build/debug/tokcat-dump cloud-probe --days 14`; per-cell resources with `--no-network --only <marker>:<variant>[:headBytes]` under `/usr/bin/time -l`.
- Endpoint shapes were read out of ChatGPT.app's `app.asar` (`/wham/usage/daily-token-usage-breakdown`, `thread_usage/query_v2`) and confirmed against the live API with the account's own credentials (read-only GETs).

### DIFF
--- a/native/LocalPackage/Sources/Collector/CloudSessions.swift
+++ b/native/LocalPackage/Sources/Collector/CloudSessions.swift
@@ -1,0 +1,312 @@
+import DataSource
+import Foundation
+
+// Cloud-session measurement surface (dev probe; not wired into the graph).
+//
+// Claude Code cloud sessions run either in Anthropic's VM or — when the
+// account has a Remote Control bridge registered — on this Mac, inside
+// `<repo>/.claude/worktrees/bridge-cse_<id>`. Only the bridged case leaves a
+// transcript under `~/.claude/projects/`, and today the graph counts it as an
+// ordinary `claude` session.
+//
+// The measurement has two axes, because both decide the answer:
+//
+//   marker   — what counts as evidence of a cloud session
+//     .pathSlug        the project directory encodes the worktree path
+//                      (`.claude/worktrees/bridge-cse_<id>`, which the slug
+//                      flattens to `-claude-worktrees-bridge-cse-<id>`)
+//     .cwdField        a row's structured `cwd` is the bridge worktree
+//     .rawSubstring    the marker bytes appear anywhere in the file
+//     .bridgeSessionID any `bridgeSessionId` whose value starts `cse_`
+//
+//   variant  — how much of each transcript is read to apply the marker
+//     .noRead          path only, zero reads
+//     .headSniff       first `headBytes`
+//     .fullScan        whole file (ground truth, and the cost ceiling)
+//
+// `.rawSubstring` and `.bridgeSessionID` are the two traps: transcripts of
+// ordinary local sessions quote the worktree path in shell output, and every
+// session the Claude app bridges locally carries a `cse_` bridge id.
+//
+// `tokcat-dump cloud-probe` renders this as JSON; the numbers in the PR come
+// from that command.
+
+/// What counts as evidence that a transcript belongs to a cloud session.
+public enum CloudMarker: String, Sendable, Codable, CaseIterable {
+    case pathSlug
+    case cwdField
+    case rawSubstring
+    case bridgeSessionID
+}
+
+/// How much of a transcript is read before the marker is applied.
+public enum CloudScanVariant: String, Sendable, Codable, CaseIterable {
+    case noRead
+    case headSniff
+    case fullScan
+}
+
+public struct CloudTranscriptReport: Sendable, Codable {
+    public var path: String
+    public var fileBytes: Int64
+    public var messages: Int
+    public var tokens: TokenBreakdown
+    public var cost: Double
+    public var dates: [String]
+}
+
+public struct CloudDayReport: Sendable, Codable {
+    public var date: String
+    public var messages: Int
+    public var tokens: Int64
+    public var cost: Double
+}
+
+public struct CloudScanReport: Sendable, Codable {
+    public var marker: CloudMarker
+    public var variant: CloudScanVariant
+    /// Bytes read per candidate file before the marker decision.
+    public var headBytes: Int
+    public var roots: [String]
+    public var filesCandidate: Int
+    /// Sum of the on-disk sizes of every candidate transcript.
+    public var candidateBytes: Int64
+    /// Bytes this (marker, variant) pair actually read.
+    public var bytesRead: Int64
+    public var wallMs: Double
+    public var transcripts: [CloudTranscriptReport]
+    public var totalMessages: Int
+    public var totalTokens: Int64
+    public var totalCost: Double
+    public var daily: [CloudDayReport]
+}
+
+public enum CloudSessionScanner {
+    /// Scan the Claude transcript roots for cloud-session transcripts.
+    ///
+    /// `home` overrides HOME so fixtures replay deterministically.
+    public static func scan(
+        marker: CloudMarker,
+        variant: CloudScanVariant,
+        headBytes: Int = 8 * 1024,
+        home: String? = nil
+    ) -> CloudScanReport {
+        let root = home ?? homeDir() ?? ""
+        let started = DispatchTime.now().uptimeNanoseconds
+        var files: [String] = []
+        var roots: [String] = []
+        for transcriptRoot in claudeTranscriptRoots(root) {
+            let found = collectFiles(transcriptRoot) { path in
+                let ext = rustExtension(path)
+                return ext == "jsonl" || ext == "json"
+            }
+            if !found.isEmpty { roots.append(transcriptRoot) }
+            files.append(contentsOf: found)
+        }
+        files.sort(by: utf8Less)
+
+        var candidateBytes: Int64 = 0
+        var bytesRead: Int64 = 0
+        var detected: [String] = []
+        for path in files {
+            candidateBytes += fileSizeBytes(path)
+            let effective = marker == .pathSlug ? CloudScanVariant.noRead : variant
+            switch effective {
+            case .noRead:
+                if pathIsCloudWorktree(path) { detected.append(path) }
+            case .headSniff:
+                let (hit, read) = fileMatches(path, marker: marker, limit: headBytes)
+                bytesRead += read
+                if hit { detected.append(path) }
+            case .fullScan:
+                let (hit, read) = fileMatches(path, marker: marker, limit: nil)
+                bytesRead += read
+                if hit { detected.append(path) }
+            }
+        }
+
+        var transcripts: [CloudTranscriptReport] = []
+        var cloudMessages: [UsageMessage] = []
+        for path in detected {
+            let messages = applyPricing(dedupMessages(parseClaudeFile(path)))
+            guard !messages.isEmpty else { continue }
+            var tokens = TokenBreakdown()
+            var cost = 0.0
+            var dates = Set<String>()
+            for message in messages {
+                tokens.addClamped(message.tokens)
+                cost += message.cost
+                dates.insert(message.date)
+            }
+            cloudMessages.append(contentsOf: messages)
+            transcripts.append(
+                CloudTranscriptReport(
+                    path: path,
+                    fileBytes: fileSizeBytes(path),
+                    messages: messages.count,
+                    tokens: tokens,
+                    cost: cost,
+                    dates: dates.sorted(by: utf8Less)))
+        }
+
+        // Same aggregation the graph would use, so the daily rows are directly
+        // comparable with `tokcat-dump graph` output.
+        let payload = buildPayload(cloudMessages)
+        let daily = payload.contributions.map {
+            CloudDayReport(
+                date: $0.date,
+                messages: Int($0.totals.messages),
+                tokens: $0.totals.tokens,
+                cost: $0.totals.cost)
+        }
+
+        return CloudScanReport(
+            marker: marker,
+            variant: variant,
+            headBytes: variant == .headSniff ? headBytes : 0,
+            roots: roots,
+            filesCandidate: files.count,
+            candidateBytes: candidateBytes,
+            bytesRead: bytesRead,
+            wallMs: Double(DispatchTime.now().uptimeNanoseconds - started) / 1_000_000.0,
+            transcripts: transcripts,
+            totalMessages: cloudMessages.count,
+            totalTokens: cloudMessages.reduce(0) { satAdd($0, $1.tokens.total) },
+            totalCost: cloudMessages.reduce(0.0) { $0 + $1.cost },
+            daily: daily)
+    }
+}
+
+// MARK: - Markers
+
+/// The bridge worktree a cloud session runs in, as it appears in a real path
+/// (`.../.claude/worktrees/bridge-cse_01TS…`). The project-directory slug
+/// flattens it to `-claude-worktrees-bridge-cse-01TS…`.
+private let cloudWorktreePathMarker = Array("worktrees/bridge-cse".utf8)
+private let cloudWorktreeSlugMarker = Array("worktrees-bridge-cse".utf8)
+private let bridgeSessionKey = Array("bridgeSessionId".utf8)
+private let cloudEnvPrefix = Array("cse_".utf8)
+/// How far past `bridgeSessionId` the `cse_` value may sit (JSON spacing slop).
+private let bridgeValueWindow = 200
+private let cwdKey = Array("\"cwd\"".utf8)
+
+/// Zero-read marker: the project directory slug encodes the cloud worktree.
+func pathIsCloudWorktree(_ path: String) -> Bool {
+    path.contains("worktrees-bridge-cse") || path.contains("worktrees/bridge-cse")
+}
+
+/// Read `limit` bytes (nil = whole file) and apply `marker`.
+func fileMatches(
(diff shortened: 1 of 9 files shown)


native/LocalPackage/Sources/Collector/CloudSessions.swift
```diff
@@ -0,0 +1,312 @@
...
+
+        return CloudScanReport(
+            marker: marker,
+            variant: variant,
+            headBytes: variant == .headSniff ? headBytes : 0,
+            roots: roots,
+            filesCandidate: files.count,
+            candidateBytes: candidateBytes,
+            bytesRead: bytesRead,
+            wallMs: Double(DispatchTime.now().uptimeNanoseconds - started) / 1_000_000.0,
```
handlecusion, line 171: `[bug]` `wallMs` is measured from `started` on line 95, so it also covers the per-detection parse loop (`parseClaudeFile` + `dedupMessages` + `applyPricing`, lines 129-152) and `buildPayload` — not just the marker pass this cell is meant to price. That makes the "probe wall" column non-comparable across rows, because a loose marker gets charged for parsing its own false positives: `bridgeSessionID` head 8 KB (1.27 s) and `cwdField` head 8 KB (0.35-0.41 s) read the same 6.7 MB, so most of that gap is the 24 extra transcripts the loose marker sent through the parser, not marker cost. Same effect on the `rawSubstring` fullScan row (10.8-11.6 s vs `cwdField` fullScan 6.8-7.0 s at identical bytes read).

Stamp a second timestamp right after the detection loop and report both:

```swift
let detectNs = DispatchTime.now().uptimeNanoseconds   // after the `for path in files` loop
...
markerWallMs: Double(detectNs - started) / 1_000_000.0,
wallMs: Double(DispatchTime.now().uptimeNanoseconds - started) / 1_000_000.0,
```

and carry `markerWallMs` through `CloudScanReport` → `MatrixRow`. The detection-only number is the one the cost comparison in the verdict table rests on.

---

native/LocalPackage/Sources/Collector/CloudSessions.swift
```diff
@@ -0,0 +1,312 @@
...
+                bytesRead += read
+                if hit { detected.append(path) }
+            }
+        }
+
+        var transcripts: [CloudTranscriptReport] = []
+        var cloudMessages: [UsageMessage] = []
+        for path in detected {
+            let messages = applyPricing(dedupMessages(parseClaudeFile(path)))
+            guard !messages.isEmpty else { continue }
```
handlecusion, line 132: `[bug]` This drops flagged transcripts that parse to zero priceable rows, and `CloudScanReport` exposes no count of the raw detections (`filesCandidate` is all candidates, `transcripts` is post-filter). So every accuracy number downstream is computed over "flagged **and** carried usage rows", not over what the marker actually flagged — `measureSourceA` takes `paths` from `report.transcripts` (CloudProbe.swift:236), so `transcripts`, `false_positives`, `precision` and `false_positive_paths` all inherit the filter.

That understates precisely what the matrix exists to expose. `~/.claude/projects` holds plenty of transcripts with no assistant usage rows (aborted sessions, summary-only files, `/compact` remnants), so `bridgeSessionID`'s real over-match is ≥ the 55 reported, and the excluded false positives never appear in `false_positive_paths` where you'd go to inspect them.

Keep the detection set on the report and score against it:

```swift
public var filesDetected: Int
public var detectedPaths: [String]
```

then compute precision / false positives over `detectedPaths` in `measureSourceA`, leaving `transcripts` as the token-attribution view. Recall is already consistent (the truth cell is filtered the same way), but it should move to the same basis so the two columns mean the same thing.

---

native/LocalPackage/Sources/Collector/CloudSessions.swift
```diff
@@ -0,0 +1,312 @@
...
+                let (hit, read) = fileMatches(path, marker: marker, limit: nil)
+                bytesRead += read
+                if hit { detected.append(path) }
+            }
+        }
+
+        var transcripts: [CloudTranscriptReport] = []
+        var cloudMessages: [UsageMessage] = []
+        for path in detected {
+            let messages = applyPricing(dedupMessages(parseClaudeFile(path)))
```
handlecusion, line 131: `[bug]` `dedupMessages` runs per file here, but the graph runs it once across every file (`UsageGraph.run` → `applyPricing(dedupMessages(messages))`) and `buildPayload` does no deduping of its own. So for any row that flags more than one transcript — `rawSubstring` fullScan (4) and `bridgeSessionID` (24 / 55) — a message that appears in two transcripts is counted twice, which is exactly what the global first-wins pass exists to prevent: resumed and forked Claude sessions do replay assistant rows with the same `requestId` into a second file, which is why `dedupKey` is global in the first place.

The single-transcript `pathSlug` / `cwdField` rows are unaffected, so the recommendation itself stands, but the 67.1 M and 406.2 M figures are inflated beyond their false-positive status, and `share` (CloudProbe.swift:271) divides a per-file-deduped numerator by a globally-deduped denominator.

Dedup once for the aggregate fields, keeping the per-file parse only for the `transcripts` breakdown:

```swift
var raw: [UsageMessage] = []
for path in detected { raw.append(contentsOf: parseClaudeFile(path)) }
let cloudMessages = applyPricing(dedupMessages(raw))
```

`totalTokens` / `totalMessages` / `totalCost` / `daily` then match what the graph would report for the same file set.

---

native/LocalPackage/Sources/Collector/CloudSessions.swift
```diff
@@ -0,0 +1,312 @@
...
+        var roots: [String] = []
+        for transcriptRoot in claudeTranscriptRoots(root) {
+            let found = collectFiles(transcriptRoot) { path in
+                let ext = rustExtension(path)
+                return ext == "jsonl" || ext == "json"
+            }
+            if !found.isEmpty { roots.append(transcriptRoot) }
+            files.append(contentsOf: found)
+        }
+        files.sort(by: utf8Less)
```
handlecusion, line 106: `[bug]` `collectFiles` already returns a sorted order, and deliberately not this one: it sorts component-wise, with `/` mapped to `0x00` (FileScan.swift:12-22), and `ClaudeParser.parse` keeps the per-root concatenation because "the downstream first-wins dedup depends on" that order. Re-sorting the merged list with `utf8Less` does two different things to it — it interleaves `projects/` and `transcripts/` into one order, and it flips pairs the collector orders the other way, since plain bytes put `-` (0x2D) before `/` (0x2F) while the collector's key puts `/` (0x00) first:

```
collector:  a/b.jsonl   then  a-x/c.jsonl
utf8Less:   a-x/c.jsonl then  a/b.jsonl
```

No numeric effect while the dedup above is per-file, but it silently diverges from the collector this probe is trying to be comparable with, and it becomes a real first-wins difference the moment that dedup goes global. Drop the line and concatenate the per-root `collectFiles` results the way `ClaudeParser.parse` does.

---

native/LocalPackage/Sources/tokcat-dump/CloudProbe.swift
```diff
@@ -0,0 +1,403 @@
...
+    process.standardOutput = stdout
+    process.standardError = stderr
+    let start = DispatchTime.now().uptimeNanoseconds
+    do {
+        try process.run()
+    } catch {
+        return nil
+    }
+    let deadline = Date().addingTimeInterval(timeout)
+    while process.isRunning && Date() < deadline {
```
handlecusion, line 525: `[bug]` The pipes are only drained at line 396, after the process has exited, so `codex` blocks as soon as it fills the OS pipe buffer (64 KB on macOS). This loop then spins to the 30 s deadline, `terminate()` fires, and `runCommand` returns `nil` — which `measureSourceB` renders as `detail: "codex not found or timed out"`. A large but perfectly good answer becomes indistinguishable from codex not being installed.

1,098 bytes for 2 tasks is far under the buffer, but `--limit 20` with long `summary` / `title` fields is the case that reaches it, and the wrong conclusion it produces ("B is unavailable") is the kind this PR is meant to settle. Drain before waiting, e.g.:

```swift
var out = Data(), err = Data()
let group = DispatchGroup()
DispatchQueue.global().async(group: group) { out = (try? stdout.fileHandleForReading.readToEnd()) ?? Data() }
DispatchQueue.global().async(group: group) { err = (try? stderr.fileHandleForReading.readToEnd()) ?? Data() }
```

keeping the deadline loop for the timeout and `group.wait()` after the process exits.

---

native/LocalPackage/Sources/tokcat-dump/main.swift
```diff
@@ -26,6 +30,7 @@ let usageText = """
 usage: tokcat-dump graph [--year YYYY] [--clients a,b]
        tokcat-dump tail-sim --dir DIR [--window-secs N] [--now-ms N]
        tokcat-dump tail-live [--ticks N] [--window-secs N]
+       tokcat-dump cloud-probe [--days N]
```
handlecusion, line 33: `[invariant]` `--only` (line 232) and `--no-network` (line 236) are accepted but missing from both this usage line and the header comment — and they are what the PR's reproduction steps ask people to run, so the per-cell numbers can't be reproduced from `tokcat-dump` with no args.

```
       tokcat-dump cloud-probe [--days N] [--only marker:variant[:headBytes]] [--no-network]
```

Same drift one level up: `AGENTS.md:17` still describes `tokcat-dump` as the parity CLI `(graph, tail-sim, tail-live)`, which is now stale by one subcommand.

---

handlecusion (commented):
🤖 **Claude Code** — 자동 리뷰

Two defects in `CloudSessionScanner.scan` move numbers that appear in the tables. `wallMs` is stamped after the detected transcripts are parsed and priced, so the "probe wall" column charges each marker for parsing its own false positives — most of the 1.27 s vs 0.35 s gap between the two head-8 KB rows is the 24 extra transcripts, not marker cost (CloudSessions.swift:171). And the accuracy columns are scored over transcripts that both matched *and* carried usage rows, so `bridgeSessionID`'s over-match is understated and the excluded false positives never reach `false_positive_paths` (CloudSessions.swift:132).

Neither overturns the verdict — `pathSlug` stays the only zero-read source that yields real tokens however the cells are timed — but the wall-clock and precision columns want a re-run before they're quoted as the basis for the decision. Third one to fix while in there: `dedupMessages` runs per file, so multi-file rows double-count against the graph's global first-wins (CloudSessions.swift:131).

Nothing here ships in the app, so nothing blocks merging. Line-level claims are from reading only — no Swift toolchain on this VM, so the timing fix needs a local re-run to confirm the new columns.



---

handlecusion:
Closing without merging.

Measured conclusion: neither cloud source can feed the daily token graph. Codex Cloud is reachable (`/wham/usage/daily-token-usage-breakdown`, 869 ms, per-day cloud/local split) but only as plan percent — the token-level per-thread endpoint (`/wham/usage/thread_usage/query_v2`) returns 403 with the CLI's OAuth token — and percent→tokens calibrates at 128K–132M tokens/percent across five days, so any conversion would be invention. Claude cloud sessions do carry real tokens, but only 1 transcript / 16,971 tokens (0.022 % of that day) exists locally against 10 fired sessions, which is not worth a graph row.

The numbers stay in this PR body, and the probe is reproducible on the branch:

    cd native/LocalPackage && swift build && ./.build/debug/tokcat-dump cloud-probe --days 14

Reopen the thread if `/wham/usage/thread_usage/query_v2` starts answering (it returns 404/403 today) — that is the Codex token source. Branch `feat/cloud-session-usage` and worktree `~/Code/tokcat-cloud-probe` are left in place.

